PR #1, opened 2021-03-19
Improve CSS only HMR by applying scoping class to every element

rixo:
Also bumps svelte-hmr to 0.14.0, to guarantee stable HMR transform code with `injectCss: false` (and `injectCss` has been changed to `false` in this version).

### DIFF
--- a/packages/vite-plugin-svelte/package.json
+++ b/packages/vite-plugin-svelte/package.json
@@ -40,7 +40,7 @@
     "require-relative": "^0.8.7",
     "slash": "^3.0.0",
     "source-map": "^0.7.3",
-    "svelte-hmr": "^0.13.0"
+    "svelte-hmr": "^0.14.0"
   },
   "peerDependencies": {
     "svelte": "^3.35.0",
--- a/packages/vite-plugin-svelte/src/utils/compile.ts
+++ b/packages/vite-plugin-svelte/src/utils/compile.ts
@@ -1,5 +1,5 @@
 import { CompileOptions, PreprocessorGroup, Processed, ResolvedOptions } from './options';
-import { compile, preprocess, walk } from 'svelte/compiler';
+import { compile, preprocess, walk, parse } from 'svelte/compiler';
 // @ts-ignore
 import { createMakeHot } from 'svelte-hmr';
 import { SvelteRequest } from './id';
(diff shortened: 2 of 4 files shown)


rixo:
Following discussion on Discord, I moved the `*{}` logic to a preprocessor.

Unfortunately, we can't use the simple `style` hook to avoid an extra parse, because it misses the cases when a `<style>` tag is added or removed from a component. For one thing, it triggers an unneeded JS update (losing state in the browser) but, in the case of removing, it also causes an error in the browser trying to fetch the inexistant virtual CSS module.

dominikg:
good catch with the style removal. do you need to parse and inject the style rule in markup though?
just append a <style> tag in markup preprocess if it does not exist (simple string search isn't foolproof but who has a string `<style` in script block but no `<style>` in markup. 

style preprocessor runs after markup and can inject the magic.

Handling removal of style nodes is a good topic to bring up too. do we need to respond with an empty style dummy module in that case? (should only ever happen if the component had style and it was deleted while devserver is running)

rixo:
> who has a string <style in script block but no <style> in markup.

[I did](https://github.com/rixo/svench/blob/87fed5cc11ede6ae57d310f0c56da619c2f663b7/svench/src/app/ExtrasPanel.svelte#L14), and it took me forever to nail it down when it bit me.

So I'm not too willing to knowingly expose others to that. This is not such an exceptional case, I believe. Code examples could probably result in that. Even more so with things like Mdsvex, probably.

Furthermore, I'm not sure the parse is so expensive. And would we really spare it by appending a `<style>` tag to the markup? Won't the compiler have to parse this modified markup, to extract the style tag anyway?

So at this point, I still believe parsing and injecting directly in markup is the most straightforward and safe solution. Maybe we can wait to see if performance becomes a pain point and reevaluate then. Even if, like I said, I don't believe there would be a lot to be scrubbed here.

> do we need to respond with an empty style dummy module in that case?

With CSS only updates and this `*{}` hack, Svelte will always produce some CSS code, and so previous styles will be cleared (which is what we want). I'm not sure it's worth the pain to detect this to try and suppress sending the empty rule to the browser in this case.

In the case where we don't have the scoping hack, though, it might be needed, but I think it shouldn't. According to tests I've done some time ago, Vite drops a stylesheet by itself when it becomes imported by nothing anymore (it had a bug where it added it back if the now-unlinked stylesheet was modified again, though). But our special logic in `handleHotUpdate` might be interfering with that somehow.

dominikg:
as discussed on discord we're going for a simpler style preprocessor first. and may revisit a more elaborate solution later

PR #2, opened 2021-03-19
Bump svelte-hmr

benmccann:
Lots of people were hitting issues with 0.13.2. I thought it might be helpful to make sure we're using a newer version


dominikg:
done

PR #13, opened 2021-03-24
feat: use `sveltePreprocess` options from vite plugins if presents

antfu:
As discussed with @dominikg.

This solves windicss compatibility problem

### DIFF
--- a/packages/vite-plugin-svelte/src/utils/preprocess.ts
+++ b/packages/vite-plugin-svelte/src/utils/preprocess.ts
@@ -109,12 +109,9 @@ export function buildExtraPreprocessors(options: ResolvedOptions, config: Resolv
 	if (options.useVitePreprocess) {
 		extraPreprocessors.push(createVitePreprocessorGroup(config, options));
 	}
-	// TODO
-	/*
-  const windiCssPlugin = config.plugins.find(p => p.name === 'vite-plugin-windicss:css');
-  if (windiCssPlugin) {
-    extraPreprocessors.unshift(createWindicssStylePreprocessorFromVite(windiCssPlugin))
-  }
-   */
+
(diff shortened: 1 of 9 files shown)


dominikg:
I fixed the lint error ( eslint plugin for svelte does not understand custom css syntax, it ignores <style lang="notcss"> but regular <stlye> is treated as css.   

I wonder if there should be a <style lang="windicss"> attribute. But thats sth that we can optionally add later.

cc @rixo @dummdidumm @kaisermann   this PR allows a pretty flexible contribution of extra svelte preprocessors by other vite plugins that use css transforms (could be useful for markup aswell).

@antfu you could try to use the markup section of the preprocessor group to scan the svelte component for classes, then you would not need an extra file read (eg not include *.svelte in the default list of scan extensions).

I'm really liking the pattern. Thank you so much for this PR

packages/vite-plugin-svelte/src/utils/preprocess.ts
```diff
@@ -109,12 +109,9 @@ export function buildExtraPreprocessors(options: ResolvedOptions, config: Resolv
...
 	}
-	// TODO
-	/*
-  const windiCssPlugin = config.plugins.find(p => p.name === 'vite-plugin-windicss:css');
-  if (windiCssPlugin) {
-    extraPreprocessors.unshift(createWindicssStylePreprocessorFromVite(windiCssPlugin))
-  }
-   */
+
+	const preprocessors = config.plugins.map(p => p && p.sveltePreprocess).filter(Boolean) as PreprocessorGroup[]
```
kaisermann, line 113:
```suggestion
	const preprocessors = config.plugins.map(p => p?.sveltePreprocess).filter(Boolean) as PreprocessorGroup[]
```

kaisermann (approved):
I'm a bit out of context but LGTM 😁

PR #16, opened 2021-03-24
feat: utilize diary and kleur

maraisr:
Hi there 👋🏻 awesome library this! Using this in a project of mine.

Just wondering how you'd feel about swapping out `debug` for `diary`? Their api's are 100% identical, [but about 20x faster](https://github.com/maraisr/diary#-benchmark).

Also swapped out chalk with kleur, which [is also about faster, and smaller](https://github.com/lukeed/kleur#performance)

--- 

Also we _could_ maybe just use diary. Alleviating the need to manually implement that `setLevel` stuff and so forth.

### DIFF
--- a/packages/vite-plugin-svelte/package.json
+++ b/packages/vite-plugin-svelte/package.json
@@ -34,8 +34,8 @@
   "homepage": "https://github.com/sveltejs/vite-plugin-svelte/tree/main/packages/vite-plugin-svelte#readme",
   "dependencies": {
     "@rollup/pluginutils": "^4.1.0",
-    "chalk": "^4.1.0",
-    "debug": "^4.3.2",
+    "kleur": "^4.1.4",
+    "diary": "^0.0.12",
     "hash-sum": "^2.0.0",
     "require-relative": "^0.8.7",
     "slash": "^3.0.0",
@@ -47,7 +47,6 @@
     "vite": "^2.1.2"
   },
   "devDependencies": {
-    "@types/debug": "^4.1.5",
     "@types/es-module-lexer": "^0.3.0",
     "@types/estree": "^0.0.46",
     "@types/hash-sum": "^1.0.0",
--- a/packages/vite-plugin-svelte/src/utils/log.ts
+++ b/packages/vite-plugin-svelte/src/utils/log.ts
@@ -1,26 +1,27 @@
 /* eslint-disable no-unused-vars */
-import chalk from 'chalk';
-import debug from 'debug';
+import kleur from 'kleur';
+import { diary } from 'diary';
 const levels: string[] = ['debug', 'info', 'warn', 'error', 'silent'];
 const prefix = 'vite-plugin-svelte';
+const { debug } = diary(`vite:${prefix}`);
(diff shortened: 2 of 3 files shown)


dominikg:
vite-plugin-svelte uses debug and chalk mainly because vite does too. So these dependencies can be shared and it integrates well with the custom rainbow colors when debugging with multiple scopes at once.

I like the idea of dropping the custom logutil / setLevel implementation. 

Ultimately i think it's in the best interest of vite-plugin-svelte to use whatever debug log system vite itself uses to stay compatible within the ecosystem. 

Vite itself exposes a logger that can be used to contribute to vites logging from plugins https://github.com/vitejs/vite/blob/7def49a49c83539b6a65c895c276fa83f7c89349/packages/vite/src/node/config.ts#L362  (one of the things i suggested a while back https://github.com/vitejs/vite/issues/501 because vite1 had the ugly habit of discarding log statements with a spinner and ultimately the reason this log util even exists bc it contained a workaround for that https://github.com/dominikg/svite/blob/master/tools/log.js#L65)

So using that may be an option too.

If diary and kleur are noticably faster and provide a benefit for vite, would it make sense for vite to use them?

cc @lukeed

lukeed:
Sharing dependencies with a primary peer generally makes sense, but in vite's case, everything is inlined...so there's no shared code here. (This wasn't always the case.) We're just bringing extra copies of XYZ right now.

> If diary and kleur are noticably faster and provide a benefit for vite, would it make sense for vite to use them?

They are significantly faster, but admittedly it may not matter when vite already isn't using the current items to their max performance either. However, what is an automatic win is their code-size reduction. Chalk is a web of 6 total packages (20kb according to bundlephobia) and kleur is 1 total package (1.9kb) – debug is 2 total packages (6kb) and diary is 1 package (1.8kb).

So while it would make sense, I generally don't suggest my packages into other projects. But if Evan is open and/or looking to implement the changes, I'm happy to help.

lukeed:
Also, in the vein of sharing code, kleur is already in use by `@sveltejs/kit` – so perhaps a way forward could be to bundle `@sveltejs/vite-plugin-svelte` into SK too so that all its chains can be shared at runtime too.

dominikg:
Hmm, doesn't inlining break independent updates? e.g. a new version of debug fixes a critical vulnerability. vite and all its users would remain vulnerable until a vite update has been released.
Same thing for new features. If vite-plugin-svelte added something and sveltekit inlined vite-plugin-svelte, sveltekit users would not be able to benefit until next release.

Not sure i like that approach. 

And while inlining takes away the benefit of shared modules i'm still leaning on sticking with debug. 
The size of a devDependency in the range of some kb isn't much of an issue and 35k ops/sek is "fast enough" for debug. 

For chalk/kleur, its a bit different. Switching to the logger provided by vite could remove the need for that altogether

maraisr:
Sorry for wasting your time @dominikg 😎

dominikg:
You did not waste my time. Sorry if my response came across blunt. Hope you still enjoy using vite-plugin-svelte.

PR #18, opened 2021-03-26
update release action and scripts

dominikg:
port pnpm caching from ci workflow
use node14
add guard for forks
update release script to call `changeset publish` in hope it does tag and generate proper changelog entries.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -49,12 +49,9 @@ jobs:
             ubuntu-latest-node-v14-pnpm-store-
 
       - name: install
-        # only run install if we don't get an exact cache hit
-        if: steps.pnpm-store.outputs.cache-hit != 'true'
         # install but don't run scripts, they could be evil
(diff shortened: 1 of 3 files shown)


dominikg:
@pngwn  what do you think? as long as changesets is in prerelease mode it should publish 1.0.0-next.x to latest but with tagging, no?

pngwn:
It will tag and create releases via the action but I'm not convinced it will use latest tags because you wouldn't normally want to tag next releases with latest if you had an existing release version. 

I think it will respect the tag version defined when you entered prerelease mode. Which I think is next in this case. The same question is being asked on the kit repo and I need to do some digging to figure out how to do it. I'll up date as soon i know more.

dominikg:
fixed a mistake i made when porting the caching. I'll try to get a vite-plugin-svelte release published this weekend. If all goes well the pattern can be applied to sveltekit too.

PR #32, opened 2021-04-15
feat: reduce use of cache in vite-plugin-svelte

dominikg:


packages/vite-plugin-svelte/src/index.ts
```diff
@@ -183,26 +183,21 @@ export default function vitePluginSvelte(inlineOptions?: Partial<Options>): Plug
...
 			const { filename, query } = svelteRequest;
-			const cachedCompileData = cache.getCompileData(svelteRequest, false);
 
 			if (query.svelte) {
-				// tagged svelte request, use cache
-				if (query.type === 'style' && cachedCompileData?.compiled?.css) {
-					log.debug(`transform returns css for ${filename}`);
-					return cachedCompileData.compiled.css;
+				if (query.type === 'style') {
+					const css = cache.getCSS(svelteRequest);
```
dominikg, line 189: This one is still a cache use, but its hard to get around it as the `code` argument contains the css in this case and we cannot compile the .svelte component again unless we get the absolute file path to it and read it ourselves, which i would really like to avoid :/

the css cache is now shared between SSR and non-SSR compile so it should always contain css even when requested via SSR injected link.

PR #35, opened 2021-04-21
Feat: async load config esm

dominikg:
cc @dummdidumm


dominikg:
for some reason running the autoprefixer-browserslist tests with jest fails to read svelte config. it works when starting it manually. :/ Investigation tomorrow. If you read this and have ideas whats happening, i'd love to hear more (;

dummdidumm:
No idea honestly, maybe Jest isn't that ESM-dynamic-imports-ready either, yet?
Btw I think it makes sense to add another test with a ESM-style config.

dominikg:
@dummdidumm added fallback to require, custom configFile option and a playground setup for good measure. 
Unfortunately the esm config cannot be loaded by jest but i validated that it starts manually. 

Would be nice if you had a quick look anyways https://github.com/sveltejs/vite-plugin-svelte/pull/35/files#diff-5fec1310cbc9274dbd46e92e8e144abbc89c90f4b8b103ef5f37f4fe6e6f15a1R13-R62

dummdidumm:
Looks good to me 👍 one more thing: you might want to add `.mjs` to the list of "try to load it via `import()`", which would enable people to use a esm config file while not setting `"type": "module"` in their package JSON (might be the case in non-SvelteKit scenarios)

Issue #29, opened 2021-04-11
Support for ESM configs


dummdidumm:
**Describe the feature**
Support loading ESM config files. Currently only CommonJS `svelte.config` files are supported because the code uses `require` to load it.

I think that [this line](https://github.com/sveltejs/vite-plugin-svelte/blob/main/packages/vite-plugin-svelte/src/utils/loadSvelteConfig.ts#L21) needs to change to use `import(..)` . Note though that TS will transpile this to a `require` anyway, that's why you need to create a `const dynamicImport = new Function('path', 'return import(path)');` function and use that one as a workaround (see https://github.com/microsoft/TypeScript/issues/43329 for more info).

**A Usecase**
Support ESM `svelte.config.js` in SvelteKit

**Alternatives**
none

**Additional context**
Blocker for https://github.com/sveltejs/kit/pull/936

dominikg:
Needs support for async config from vite, this has been merged but not released https://github.com/vitejs/vite/pull/2758

Also does this mean that going forward there will be no svelte.config.cjs and we should discourage its use or are both allowed?

And while we're at it, whats the list of locations vite-plugin-svelte should look for svelte.config.c?js ?  (see https://github.com/sveltejs/vite-plugin-svelte/issues/23 for context)

dominikg:
Update, async config hook might be a better place: https://github.com/vitejs/vite/pull/2800

dummdidumm:
Both `cjs` and `js` should be supported. It should check `.js` first, then `.cjs` to be in line with the order how language-tools and SvelteKit (will) resolve it.

Regarding "where to look": I think this was never specified. But we should, at some point, to save us some maintenance 😄

dominikg:
Initial code here https://github.com/sveltejs/vite-plugin-svelte/compare/feat/async-load-config-esm  needs to wait until vite 2.16 is released and peer dependency updated.

should we try to get rid of everything non-esm and make vite-plugin-svelte itself a type: module too? 

Here's another require call: https://github.com/sveltejs/vite-plugin-svelte/blob/a2acab981c91dd47aa036798ef0bf79d9c805342/packages/vite-plugin-svelte/src/index.ts#L161 , anything else?

dummdidumm:
You also need to replace the `require-relative` dependency I think.

janosh:
Very interested in this PR. Anything I can do to help?

dominikg:
[...]

PR #38, opened 2021-04-28
fix: use pathToFileURL to ensure windows compatibility

dominikg:
see https://github.com/sveltejs/vite-plugin-svelte/issues/29#issuecomment-828218723

cc @dummdidumm


dummdidumm (approved):
tested this locally and it works.
PS: changeset 😄

PR #45, opened 2021-05-17
Fix/compiler warnings

dominikg:
compiler warnings were not logged because plugin-context wasn't passed to compile call. this fixes it


dominikg:
unfortunately the handleHotUpdate hook of vite does not offer PluginContext as "this".  so using  'warn' isn't going to work there. 

Either provide a custom 'onwarn' option that delegates back to vite somehow or completely handle it ourselves?

Issue #47, opened 2021-05-26
"dynamicImportDefault" doesnot work as expected


SOVLOOKUP:
## Description
I am using svelte kit with [midway/hooks](https://github.com/midwayjs/hooks) which needs a vite plugin to work.

But when importing the plugin in `svelte.config.js`, ["dynamicImportDefault" function](https://github.com/sveltejs/vite-plugin-svelte/blob/main/packages/vite-plugin-svelte/src/utils/loadSvelteConfig.ts#L12) will fail with err:

`> A dynamic import callback was not specified.`

I found the problem is:
Dynamic `import` can only work in ES module file. So after tsc compiling, ["dynamicImportDefault" function](https://github.com/sveltejs/vite-plugin-svelte/blob/main/packages/vite-plugin-svelte/src/utils/loadSvelteConfig.ts#L12) won't work as expected in `dist/index.js`(a cjs file). 

## Reproduction

https://github.com/SOVLOOKUP/midway-svelte

dominikg:
Thank you for reporting this issue. 
Is it possible that you have to push your latest code to the repo? It seems there are dependencies missing in package.json.
After i installed them and changed `hooks()` to `hooks.default()` in svelte.config.js i got the following error
```
[vite-plugin-svelte] failed to require config /home/dominikg/develop/reproductions/midway-svelte/svelte.config.js
Error [ERR_REQUIRE_ESM]: Must use import to load ES Module: /home/dominikg/develop/reproductions/midway-svelte/svelte.config.js
require() of ES modules is not supported.
require() of /home/dominikg/develop/reproductions/midway-svelte/svelte.config.js from /home/dominikg/develop/reproductions/midway-svelte/node_modules/.pnpm/@sveltejs+vite-plugin-svelte@1.0.0-next.10_svelte@3.38.2+vite@2.3.4/node_modules/@sveltejs/vite-plugin-svelte/dist/index.js is an ES module file as it is a .js file whose nearest parent package.json contains "type": "module" which defines all .js files in that package scope as ES modules.
Instead rename svelte.config.js to end in .cjs, change the requiring code to use import(), or remove "type": "module" from /home/dominikg/develop/reproductions/midway-svelte/package.json.

    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1080:13)
    at Module.load (internal/modules/cjs/loader.js:928:32)
    at Function.Module._load (internal/modules/cjs/loader.js:769:14)
[...]

PR #49, opened 2021-05-27
fix: "fix: dynamicImport svelte config #47"

SOVLOOKUP:
I solved this bug #47 by refacting the project to esm.

This change must use `tsup` to build, esbuild cannot deal with some compatibility problems between cjs and esm, so I replace it.

you can verify it with running https://github.com/SOVLOOKUP/midway-svelte


eikaramba:
can this PR be merged as it currently prevents anyone from using latest svelte-kit with windicss (using vite adapter) ? not want to be annoying, just asking...

dominikg:
Thank you for starting this, unfortunately it required additional changes for testing. I went ahead and implemented it here https://github.com/sveltejs/vite-plugin-svelte/pull/54  please let me know if that works for you.

PR #50, opened 2021-06-03
Svg support

bluwy:
Add support for importing `.svg` files as Svelte components when using `extensions: ['.svelte', '.svg']`.

## Changes

1. `vite-plugin-svelte` will load SVGs from the filesystem if `.svg` is specified in `extensions`.
2. Added new `asset-import` playground to test custom extension Svelte components in general, currently `.svg`.
3. Also added test in `asset-import` playground.

## Notes

1. [This is how](https://github.com/vitejs/vite/blob/dfab3ed82a4651a323542cd5429b281272ae8ec6/packages/vite/src/node/plugins/asset.ts#L64-L68) Vite handles importing SVGs raw currently by file path. Notably, it searches for a potential public file and/or clean the URL, but I don't think we need to do either of those if an SVG is assumed to be a Svelte file.
2. I've removed `.html` imports for the `asset-import` playground since there's an issue that the plugin would compile the entrypoint HTML too (bad idea).


packages/playground/asset-import/vite.config.js
```diff
@@ -0,0 +1,12 @@
+const svelte = require('@sveltejs/vite-plugin-svelte');
+const { defineConfig } = require('vite');
+
+module.exports = defineConfig(({ command, mode }) => {
+	const isProduction = mode === 'production';
+	return {
+		plugins: [svelte({ extensions: ['.svelte', '.html', '.svg'] })],
```
dominikg, line 7: at least in vite1, .html extension didn't work for svelte, vite treats that entirely different. Does it work now?

dominikg:
cc @pngwn  

i think it may be better to use `.svelte.svg`  and make the extension configurable to allow both regular svg assets and "svelty" ones. In general everything that doesn't have to run through our plugin helps performance.

dominikg:
another idea would be fully automatic handling. if filter matches and file ends with known asset ext, load it in load hook to prrevent vite asset handling.

packages/playground/asset-import/vite.config.js
bluwy: Yeah, it seems to work. Though I should probably test with a more sophisticated Svelte code in the HTML to check if Vite is doing anything else.

Regarding the failing tests, I had issues getting Playwright to startup the browser so I couldn't actually test locally. So I'm sort of winging it for this PR.

bluwy:
The latter idea resembles more to the current PR implementation. I tried to keep the change as small as possible. 

Another idea I had was to introduce a special `?svelte` query when importing from a path, which the plugin will handle by default. Could be a bit overboard.

packages/playground/asset-import/vite.config.js
bluwy: @dominikg Through some testing, the current change can handle HTML inports only in dev mode, but fails in build mode. Turns out setting `.html` extension would trigger the plugin to compile the entrypoint HTML as a Svelte component, which generates a final empty bundle.

I've removed HTML support for now, but there could be ways to support this in the future, like detecting HTML files to avoid Svelte compile via the `transformIndexHtml` hook, though a bit sketchy. Would need to figure out a more solid strategy for that.

I also fixed the tests.

bluwy:
Closing this now as the strategy used no longer works in the latest version of Vite (not sure why yet). Will open another PR with a proper implementation.

PR #53, opened 2021-06-09
Add docs on importing third-party Svelte libraries

bluwy:
Am guilty of knowing this trick, and not announcing this sooner. Alas I pulled the plug :smile: 

Added a section about using `optimizeDeps.exclude`. I know there are plans to remove this caveat soon, but I'm not sure when. In the mean time, it would be great to document this so people don't get frustrated by Svelte's ecosystem.

Feel free to edit the docs!

Note: We could have this info in SvelteKit's FAQ too.


dominikg:
thank you so much :heart_eyes:

PR #54, opened 2021-06-09
Feat/esm

dominikg:
conversion to esm. superseeds https://github.com/sveltejs/vite-plugin-svelte/pull/49


packages/playground/package-json-svelte-field/package.json
```diff
@@ -0,0 +1,18 @@
+{
+  "name": "playground-package-json-svelte-field",
+	"private": true,
+  "version": "1.0.0",
+  "scripts": {
+    "build": "vite build",
+    "dev": "vite"
+  },
+	"dependencies": {
```
benmccann, line 9: Spacing is inconsistent in this file

packages/vite-plugin-svelte/tsconfig.json
```diff
@@ -1,6 +1,8 @@
 {
   "include": ["src"],
-  "exclude": ["**/*.spec.ts"],
+  "exclude": [
+		"**/*.spec.ts"
+	],
```
benmccann, line 5: Alignment looks a bit off here

.changeset/happy-toys-boil.md
```diff
@@ -0,0 +1,15 @@
...
+feat: convert to es module and provide cjs fallback
+
+If you are using vite-plugin-svelte with require, you should switch to esm and use import.
+An example can be found in the usage section of the [readme](README.md)
+
+continuing with cjs/require is discouraged but if you must use it, update your require statement to add `.default`
+
+```diff
+- const svelte = require('@sveltejs/vite-plugin-svelte');
++ const svelte = require('@sveltejs/vite-plugin-svelte').default;
```
antony, line 14: is

```
const { default: svelte } = require('@sveltejs/vite-plugin-svelte');
```

easier on the eye?

.gitignore
```diff
@@ -1,22 +1,45 @@
-.DS_Store
```
antony, line 1: This seems a bit like PR pollution - is it necessary to change all of this?

package.json
```diff
@@ -15,9 +15,9 @@
     "test:ci:serve": "cross-env VITE_PRESERVE_BUILD_ARTIFACTS=1 jest --verbose --no-cache --runInBand --force-exit --ci --json --outputFile=\"temp/serve/jest-results.json\"",
     "test:ci:build": "cross-env VITE_TEST_BUILD=1 VITE_PRESERVE_BUILD_ARTIFACTS=1 jest --verbose --no-cache --runInBand --force-exit --ci --json --outputFile=\"temp/build/jest-results.json\"",
     "lint": "eslint --ignore-path .gitignore '**/*.{js,ts,svelte,html,svx,md}'",
-    "lint:fix": "pnpm lint -- --fix",
+    "lint:fix": "pnpm run lint -- --fix",
```
antony, line 18: I think pnpm added a feature recently to allow you to drop the `run` part. Either way I feel like this bloats the PR - makes it harder to review.

packages/playground/autoprefixer-browerslist/vite.config.js
```diff
@@ -1,4 +1,4 @@
-const svelte = require('@sveltejs/vite-plugin-svelte');
+const svelte = require('@sveltejs/vite-plugin-svelte').default;
```
antony, line 1: as per readme, but it's a style choice I guess.

packages/playground/default-svelte-template/vite.config.js
```diff
@@ -1,4 +1,4 @@
-const svelte = require('@sveltejs/vite-plugin-svelte');
+const svelte = require('@sveltejs/vite-plugin-svelte').default;
```
antony, line 1: it has occurred to me that since we say "if you must use cjs" and then use it in all of our examples, they're not great examples... :)

packages/vite-plugin-svelte/src/handle-hot-update.ts
```diff
@@ -2,7 +2,7 @@ import { ModuleNode, HmrContext } from 'vite';
 import { Code, CompileData } from './utils/compile';
 import { log, logCompilerWarnings } from './utils/log';
 import { SvelteRequest } from './utils/id';
-import { VitePluginSvelteCache } from './utils/VitePluginSvelteCache';
+import { VitePluginSvelteCache } from './utils/vite-plugin-svelte-cache';
```
antony, line 5: :heart:

antony (approved):
only nitpicks

lukeed (commented):
If you use a named export you can skip the `require('@sveltejs/vite-plugin-svelte').default` stuff. 

And since you're adding `"exports"` map support _and_ `type: module`, this is already a breaking change, so might as well lean into it further and add the named export. The benefit of this is that it's more consistent for the user, and the type definitions will work for both ESM _and_ CJS

bluwy:
My 2c. I'd personally prefer if we stick with default exports in contrary to @lukeed's suggestion, to be consistent with Vite's plugin ecosystem. `require('@sveltejs/vite-plugin-svelte').default` is still negligible, plus most Vite configs are written in ESM instead of commonjs, so it shouldn't affect too many projects.

On that topic, I also agree with @antony that we should encourage ESM in the Vite config files.

.changeset/happy-toys-boil.md
dominikg: got a slight preference for the former hence i went with it. But if you feel the {default: svelte} syntax is more modern or has other benefits i'm open for changing it. 

cjs is a fallback anyways and i'd love to make a total switch. more on that below....

package.json
dominikg: that feature made me add it back, i wasn't sure if it would prefer a `lint` dependency  over the lint script.

packages/playground/default-svelte-template/vite.config.js
dominikg: The thing is, these are not only examples but also e2e testbeds. and jest still isn't playing ball even with 27 and initial esm support. Gotta spend more time to convert the whole setup and stuff like https://github.com/facebook/jest/issues/11438 makes me hesitant to even start on it

dominikg:
Thank you for all your comments. Greatly appreciated :heart: 

I like @lukeed suggestion of a named export and had thought about it too, but ultimately went with the default setup from other vite plugins @bluwy mentioned.

Sharable typings are a valid reason to do it and maybe we can start a trend for other vite plugins to follow?
What should the export be named though? `svelte` is the current name used in configs and examples, any objections?

I'd love to ditch cjs completely but feel thats a bit too soon. Esp. because right now it is still needed for tests with jest.

Regarding more prominent esm examples, should the playgroud be spit into e2e and examples section?
Gotta admit its a bit messy right now.

.changeset/happy-toys-boil.md
dominikg: named export solved this

.gitignore
dominikg: yes, i got carried away with some cleanup, sorry.

dominikg:
updated to named export but still included a default export with a deprecation warning:
https://github.com/sveltejs/vite-plugin-svelte/pull/54/files#diff-71fd6d8bccf1f224607bb27c94b502d32abf70103f0a2a59ad3acb5d229c443bR228

Without the default export, sveltekit breaks until updated (which includes the test in playground/kit-node) 
We should coordinate a release of sveltekit  that uses the named export so users can update gracefully as soon as they encounter the warning.

dominikg:
hmm, asked about it on vite discord and while most plugins use a default export, there is eg. vite-plugin-pwa that uses a named export.
The examples on vite docs use a default export ( https://vitejs.dev/guide/api-plugin.html#transforming-custom-file-types ) as do all the official rollup plugins.

here are the links to the discussions back at terser plugin:
https://github.com/TrySound/rollup-plugin-terser/issues/31 https://github.com/TrySound/rollup-plugin-terser/issues/7

if it is technically more consistent i don't care much about two additional {} in an import statement, but it seems others do care and i'd like to avoid having discussions about this in the future, so ideally we can decide on a way forward here and stick to it.

so what should it be?

1)  default export
- less work with sveltekit transition
- follows unwritten convention

2) named export 
- improved cjs experience
- needs work to ensure smooth transition for sveltekit users

@lukeed  maybe you can explain more about the benefits of the named export for cjs.

benmccann:
Another option would be to export as named and default export, but only put the named export in the type definitions and docs and make that the preferred but not required way

lukeed:
TypeScript only supports 1 `export` format in a definition file. Using a named export is the only way to expose information across ESM/CJS access. Specifically, `export default` _excludes_ `module.exports` and this is why `require('foo').default` happens, even though it's _not_ the CommonJS equivalent – it's a forced outcome of prior tooling mistakes & we're just keeping it around because we, as an ecosystem, think that's okay/how it's supposed to be. The use of `.default` means that the CJS source has to look like `module.exports.default = ` or `exports.default = `, which is silly.

@TrySound cares about this too & that's why `rollup-plugin-terser` often looks like the odd one out. He's correct, though – everyone else is either pure ESM, doesn't care about CJS access, or expects Rollup users to use `import` & rely on Rollup to do the pre-bundle interop/conversion for them.

bluwy:
I guess it comes down to whether we want an awkward syntax for ESM or CJS. Given a majority of our userbase is on ESM (since the default Vite starter config is in ESM), I'd prefer default exports to please the crowd.

If it all comes down to preference, perhaps a public poll could end this once and for all.

dominikg:
removed the default export completely (as expected this caused the kit-node tests to fail). 
With a coordinated release of vite-plugin-svelte and sveltekit, users of sveltekit should not run into it (unless they update vite-plugin-svelte without updating kit).

PR #68, opened 2021-06-23
feat: allow svelte imports in optimizeDeps include

dominikg:
cc @bluwy

### DIFF
--- a/packages/vite-plugin-svelte/src/utils/options.ts
+++ b/packages/vite-plugin-svelte/src/utils/options.ts
@@ -2,6 +2,7 @@
 import { ConfigEnv, UserConfig, ViteDevServer } from 'vite';
 import { log } from './log';
 import { loadSvelteConfig } from './load-svelte-config';
+import { SVELTE_HMR_IMPORTS, SVELTE_IMPORTS, SVELTE_RESOLVE_MAIN_FIELDS } from './constants';
 
 const knownOptions = new Set([
 	'configFile',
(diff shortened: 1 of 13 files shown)


packages/vite-plugin-svelte/src/utils/options.ts
```diff
@@ -151,6 +152,45 @@ export async function resolveOptions(
...
+export function buildExtraViteConfig(
+	options: ResolvedOptions,
+	config: UserConfig
+): Partial<UserConfig> {
+	const allSvelteImports = [...SVELTE_IMPORTS, ...SVELTE_HMR_IMPORTS];
+
+	// exclude svelte imports from optimization unless explicitly included
+	const excludeFromOptimize = allSvelteImports.filter(
+		(x) => !config.optimizeDeps?.include?.includes(x)
+	);
```
bluwy, line 164: Something I just tested, if the user provide the svelte imports in `optimizeDeps.include` of the config directly, meaning the svelte imports would end up in both `optimizeDeps.include` and `optimizeDeps.exclude`, Vite pre-bundling seems to have `include` take precedence. But I think it's safer this way too.

Just something I found out.

bluwy (approved):
Thanks for handling this. Appreciate it!!

packages/vite-plugin-svelte/src/utils/options.ts
dominikg: yeah, even if include takes precedence, having it in both is confusing.

PR #72, opened 2021-06-25
feat: restart server when svelte config file changes (fixes #63)

### LINKED ISSUE #63: Restart Vite server on svelte config file change
**Describe the feature**
When updating `svelte.config.js`, it will restart the Vite dev server,

**A Usecase**
Convenient if this is handled automatically.

**Alternatives**
Manually restart dev server.

**Additional context**
This is how [vite-plugin-windicss handles](https://github.com/windicss/vite-plugin-windicss/blob/6324d294db73a6033bee407bca8c4610dc87f404/packages/vite-plugin-windicss/src/index.ts#L86-L94) currently.

I think it would be a simple addition, but I have concerns if this could negatively affect SvelteKit.

dominikg:
the dynamic import with ?t= param makes the workaround even more ugly but otherwise a cache might return the old config instead of the changed one.

Tests needed, esp. to ensure that output changes due to different compiler settings are loaded in the browser. 

Idea: test virtual css module exists, then switch emitCss off, test that page reloads and virtual module no longer exists.


bluwy:
> the dynamic import with ?t= param makes the workaround even more ugly but otherwise a cache might return the old config instead of the changed one.

Looks like Vite is doing [something similar](https://github.com/vitejs/vite/blob/7362e6e9a7a0f0177b467f1cf80552acf22c46a0/packages/vite/src/node/config.ts#L762-L763) too, so I don't think it's all too ugly.

Really like the changes here, looks cleaner than vite-plugin-restart's implementation. Gonna test this locally now.

dominikg:
ahhh, great. was wondering about how to clear require cache https://github.com/vitejs/vite/blob/7362e6e9a7a0f0177b467f1cf80552acf22c46a0/packages/vite/src/node/config.ts#L772

the mjs detection might be worth using too

packages/vite-plugin-svelte/src/utils/watch.ts
```diff
@@ -33,9 +41,38 @@ export function setupWatchers(
...
+	const possibleSvelteConfigs = knownSvelteConfigNames.map((cfg) => path.join(root, cfg));
+	const restartOnConfigAdd = (filename: string) => {
+		if (possibleSvelteConfigs.includes(filename)) {
+			log.info(`svelte config added: restarting vite server. - file: ${filename}`);
+			server.watcher.emit('change', viteConfig);
+		}
+	};
+
+	const restartOnConfigChange = (filename: string) => {
+		if (filename === svelteConfigFile) {
```
bluwy, line 53: bug: `svelteConfigFile` here needs to be an absolute path to possibly match this condition, otherwise custom config files doesn't work.

dominikg: https://github.com/sveltejs/vite-plugin-svelte/blob/4e876d2c61f6f8b009618aa84c7b3bdd4bc30d55/packages/vite-plugin-svelte/src/utils/load-svelte-config.ts#L73

should be absolute, the path is resolved and updated while loading

bluwy: It isn't absolute for `packages/vite-plugin-svelte/src/utils/watch.ts` though, through some logging the value was always e.g. `svelte.config.custom.cjs`. I'm currently testing it at the `configfile-custom` e2e test package, and the config change doesn't trigger a server restart.

dominikg:
finally! the refactored e2e setup means vite is no longer run inline during tests but instead via execution of package.json dev or build + preview scripts.  (with the exception of vite-ssr where the custom setup still needs conversion). Because the project setup no longer relies on jest to start vite, we can also use esm projects for e2e tests :tada:

PR #101, opened 2021-07-15
feat: generate preprocessor sourcemaps

dominikg:
some community provided preprocessors do not return sourcemaps for their transforms which can be a source for errors during hmr.


packages/vite-plugin-svelte/src/utils/sourcemap.ts
```diff
@@ -0,0 +1,44 @@
+import MagicString, { MagicStringOptions } from 'magic-string';
+import { diff_match_patch, DIFF_DELETE, DIFF_INSERT } from 'diff-match-patch';
+
+export function buildMagicString(
```
bluwy, line 4: Pretty neat utility, I wonder if there's an external library that does exactly this :D

PR #105, opened 2021-07-20
feat: import svelte types

### LINKED ISSUE #10: Import svelte types instead of duplicating them
**Describe the feature**
Currently preprocess and compilerOptions are duplicated in this repo. Import them from svelte instead

**A Usecase**
The types should always reflect what the svelte version in use has to offer

**Additional context**

bluwy:
Fixes #10

I initially wanted to import the types from svelte directly for every other files, but eslint keeps complaining of `node/no-missing-import`, and the imports are *ugly* :D So I've only modified the `option.ts` file, which re-exports the types needed. Though, let me know if we do want to import from svelte directly, and we could try that path.

I also removed `--dts-resolve` from `tsup`. It was unable to resolve `estree`, likely from Svelte compiler's dependencies. Removing that doesn't change the final dts however, so it looks safe.

### DIFF
--- a/.changeset/famous-lemons-draw.md
+++ b/.changeset/famous-lemons-draw.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/vite-plugin-svelte': patch
+---
+
+import svelte types instead of duplicating them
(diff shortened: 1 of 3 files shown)


bluwy:
Forgot to mark the exported types with `export type` which caused these imports to be included in the JS bundle before. Fixed.

dominikg:
does this actually work? svelte does not export the types, hence the eslint warnings. I thought the first step for this was to make these types public in svelte.

bluwy:
Yeah it does work. I'm guessing eslint wasn't scanning ts files, hence the errors. They're deep imports though, but it's the only way to import the types, so they sort-off become public APIs (see also https://github.com/sveltejs/svelte/pull/6100). I also did this in the past and it works well.

PR #107, opened 2021-07-21
chore: replace querystring with URLSearchParams

### LINKED ISSUE #106: Replace querystring with URLSearchParams
### Describe the problem

It is recommended to avoid using querystring. I'm getting a warning when building SvelteKit about querystring and can't figure out where it's coming from. This is one of a couple libraries that use it. I'm hoping if we can remove it then I won't get the warning anymore

https://github.com/sveltejs/vite-plugin-svelte/blob/d1d2638b247830852faa89e7b9bc9a430b81ba51/packages/vite-plugin-svelte/src/utils/id.ts#L2

Use URLSearchParams instead

### Alternatives considered

status quo

### Importance

nice to have

bluwy:
Fixes #106 

Not sure of the performance implications, but it seems nice to avoid soon-to-be deprecated APIs.

### DIFF
--- a/.changeset/breezy-kids-divide.md
+++ b/.changeset/breezy-kids-divide.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/vite-plugin-svelte': patch
+---
+
+replace querystring with URLSearchParams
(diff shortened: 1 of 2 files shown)


dominikg:
thanks, that looks very clean. :+1:   

I think `querystring` needs to be removed from package.json and the lockfile aswell `pnpm remove querystring` in `packages/vite-plugin-svelte` should do

Performance wise it may have a small impact as its used in multiple hooks, but one less external dependency is a plus too.

bluwy:
Didn't notice it was part of the dependencies, I thought that was imported from Node's builtin modules. Will remove that too.

benmccann:
Thanks!!

Issue #121, opened 2021-07-26
Improve compile performance for ssr


dominikg:
### Describe the problem

Components are compiled twice for ssr.

1. with `generate: "ssr"` for the serverside render
2. with  `generate: "dom"` for the clientside bundle

The current implementation runs preprocess for each step separately ​which is wastful, especially for more demanding preprocessors like postcss or mdsvex.

Also components that are never compiled for ssr still get compiled with hydratable: true. Even though they are clientside only.

### Describe the proposed solution

To avoid duplicate preprocessing, compile both variants in parallel during transform with `ssr: true` and cache + reuse the compiled dom variant in subsequent transform with `ssr: false`

Only set `hydratable: true` in this parallel compilation, components that are only transformed with `ssr: false` can be compiled with `hydratable: false` to reduce output size.

**Possible issue:** 
the special `svelte/ssr` resolve mechanism could break the dom output if it was compiled during the transform with `ssr: true`

If that is indeed the case, the preprocess result could be cached instead and used as an indicator that the client variant needs to be compiled with `hydratable: true`

### Alternatives considered

stick with the current implementation. This plugin has had issues with aggressive caching for ssr before and caching was completely removed to avoid it.

### Importance

nice to have

dominikg:
Addendum: Sveltekit allows to disable ssr for individual pages or in general https://kit.svelte.dev/docs#ssr-and-javascript

Should vite-plugin-svelte try to read kit config too( svelteconfig.kit namespace ) or should kit be responsible to setup the plugin accordingly?

dominikg:
After talking about this with @rixo i realized that it would be very hard to implement within vite-plugin-svelte, so instead of trying to be smart about it a new experimental option to allow dynamic modification of compileOptions is going to be added, see #122 

```js
	dynamicCompileOptions?: (data: {
		filename: string; 
		code: string; /* preprocessed code */
		compileOptions: Partial<CompileOptions>; /* compileOptions to be used for compile*/
	}) => Promise<Partial<CompileOptions> | void> | Partial<CompileOptions> | void; /* return partial with wanted modifications or void to use unmodified */
```

kit could use this to identify components that are never serverside rendered and set hydratable return `{hydratable: false}`

benmccann:
I don't quite understand everything posted above, but if we only have SSR or only have CSR then hopefully we're only compiling the components once. If we have both, then caching the preprocessing step sounds useful.

dominikg:
[...]

Issue #120, opened 2021-07-26
Improve handling of compilerOptions.hydratable


dominikg:
### Describe the problem

Currently vite-plugin-svelte defaults  to `hydratable: true` which is not the default for svelte and mostly due to the reason that it reliably works with and without ssr.

In situations where you don't use ssr and/or don't need clientside hydration, this leads to increased output size.

### Describe the proposed solution

Only compile with `hydratable: true` when one of the following is true

* compilerOptions.hydratable is set to true in svelte config
* compilerOptions.hydratable is not defined in svelte config and viteConfig.ssr exists 

### Alternatives considered

stick with the current default. Users that want `hydratable: false` should set it in svelte config.

### Importance

nice to have

benmccann:
Sounds fine to me as long as we specify a default of true in SvelteKit still

dominikg:
Explicit compilerOptions.hydratable in kit is a good idea if #121 doesn't work out. 

It would work even with the current version of kit as it sets `viteConfig.ssr`.

That check isn't ideal (false positive if the config value is set but ssr isn't used), but the more correct/strict check
```js
const isSSR = !!(viteEnv.command === 'serve' ? viteConfig.server.middlewareMode : viteConfig.build.ssr);
``` 
 would be false for the build of client bundle, where hydratable is needed.

bluwy:
I never realized `hydratable` was set to `true` by default all these time. Is it true for production builds too? I think this is a welcomed change.

dominikg:
PR in kit to set hydratable: https://github.com/sveltejs/kit/pull/2024

using viteConfig.ssr turned out to be not the best idea as this is not always set when hydratable is needed. So instead of adding an automatic default that only works half of the time, vite-plugin-svelte will default to `hydratable: false` always

dominikg:
released https://github.com/sveltejs/vite-plugin-svelte/blob/main/packages/vite-plugin-svelte/CHANGELOG.md#100-next15

PR #122, opened 2021-07-27
feat: default to hydratable: false and add dynamicCompileOptions

dominikg:
see #120 and #121


.changeset/calm-dancers-accept.md
```diff
@@ -0,0 +1,9 @@
+---
+'@sveltejs/vite-plugin-svelte': major
+---
+
+change default value of compilerOptions.hydratable to false
+
+This is done to align with svelte compiler defaults and improve output in non-ssr scenarios.
+
+Add `{compilerOptions: {hydratable: true}}` to vite-plugin-svelte config if you need hydration (eg. for ssr)
```
rixo, line 9: Should mention that defaults to true when Vite options.ssr is true, no? Or isn't this the case?

dominikg: not the case as not all SSR solutions set that and I chose consistency over automatic default that only works some of the time

dominikg:
After https://github.com/sveltejs/kit/pull/2024 has been merged and released, update kit version here and tests should pass.

PR #128, opened 2021-08-03
DO_NOT_MERGE update vite to 2.5.0-beta.0 to check ci

dominikg:
lets see if something breaks


dominikg:
2.5 has been released

PR #133, opened 2021-08-05
ci: refactor ci workflow to reduce number of jobs

dominikg:
changes

- remove the 3 pnpm store jobs and delegate to setup-node
- join build,lint,audit
- reduce test matrix from 3x3 to 3+1+1
- only run test matrix if build check was successful

num jobs | total  |  concurrent
------------ | ------------- | -------------
before | 15 | 12
after | 6 | 5

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -42,10 +42,10 @@ jobs:
         run: pnpm run build
       - name: lint
         if: (${{ success() }} || ${{ failure() }})
-        run: pnpm run lint && exit 1
+        run: pnpm run lint
       - name: audit
         if: (${{ success() }} || ${{ failure() }})
-        run: pnpm audit && exit 1
+        run: pnpm audit
 
   # this is the test matrix, it runs with node14 on linux,windows,macos + node12,16 on linux
   # it is skipped if the build step of the checks job wasn't successful (still runs if lint or audit fail)
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,8 +17,9 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ ubuntu-latest ]
-        node: [ 14 ]
+        # pseudo-matrix for convenience, NEVER use more than a single combination
+        node: [14]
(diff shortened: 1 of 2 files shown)


dominikg:
surprisingly pnpm caching doesn't even have that big of an effect on runtime. first run was 3:45, second now 3:26 
the bigger matrix before was as fast sometimes, but stalling mac/win sometimes slowed it down, so all in all the average time should also go down.

benmccann:
nice!! seems like a good change

.github/workflows/ci.yml
```diff
@@ -13,159 +10,78 @@ on:
...
+    if: (${{ success() }} || ${{ failure() }}) && (${{ needs.checks.output.build_successful }})
     timeout-minutes: 10
     runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
       matrix:
-        os: [ ubuntu-latest, windows-latest, macos-latest ]
-        node: [ 12, 14, 16 ]
+        node-version: [14]
+        os: [ubuntu-latest, macOS-latest, windows-latest]
```
bluwy, line 60:
```suggestion
        os: [ubuntu-latest, macos-latest, windows-latest]
```

Would be nice to have macos small caps so github makes a cleaner output :grin:

.github/workflows/ci.yml
```diff
@@ -13,159 +10,78 @@ on:
...
+  checks:
     timeout-minutes: 5
     runs-on: ${{ matrix.os }}
-
     strategy:
-      fail-fast: false
       matrix:
-        os: [ ubuntu-latest, windows-latest, macos-latest ]
+        os: [ ubuntu-latest ]
         node: [ 14 ]
```
bluwy, line 21:
```suggestion
        node: [ 14 ]
        os: [ ubuntu-latest ]
```

Flipping this would make github action's output a bit consistent :grin: 

image

bluwy (approved):

.github/workflows/ci.yml
dominikg: done

.github/workflows/ci.yml
dominikg: done

Issue #125, opened 2021-08-02
Automatically add svelte libraries to vite.optimizeDeps.exclude


dominikg:
### Describe the problem

dedupe does not work for optimized dependencies and this breaks stuff.

See 
https://github.com/sveltejs/vite-plugin-svelte/issues/124
https://github.com/vitejs/vite/issues/3910
https://github.com/bluwy/vite-svelte-dedupe

### Describe the proposed solution

To avoid this problem, automatically add svelte libraries to optimizeDeps.exclude.

Finding those libraries isn't bulletproof, the best you can do is check for the existance of a "svelte" field in it's package.json,

see https://github.com/sveltejs/kit/blob/69b92ec16ce92b86bc5a540daa8ddbd84ba400ef/packages/kit/src/core/utils.js#L85

### Alternatives considered

keep the status quo and users have to apply it manually

### Importance

would make my life easier

mattjennings:
I like this. It's worth noting that svelte-kit package does not add a pkg.svelte property which is unfortunate, but it does add the .svelte files to pkg.exports. Maybe we could check those as well?

dominikg:
it does? then it would also have to check that field in the code I linked 🤔 
thanks for pointing it out.

also vite-plugin-svelte and other bundler plugins would also need to make sure to resolve via that custom ext.

mattjennings:
That would explain why my libraries using kit package only work when under `dependencies`. Otherwise, they get that `Must use import to load ES Module` message unless it's added to `vite.ssr.noExternal` manually

For reference, here's the package.json: https://unpkg.com/svelte-modals@1.0.3/package.json.

dominikg:
ah, it's only adding the path mapping, not a custom field. that's OK.

your problem is likely the incompatibility with esm-only packages in vite SSR. 
https://github.com/vitejs/vite/discussions/4230#discussioncomment-1011534

mattjennings:
Oh, yes, sorry, I meant to say we could skim through pkg.exports with .svelte extensions as an alternative to looking for pkg.svelte if it's not there. There is an [issue being discussed](https://github.com/sveltejs/kit/issues/1959) for kit to add pkg.svelte, so maybe this is more of a kit issue anyway.

Thanks for the heads-up on the esm-only SSR bug though, that looks like the culprit.

benmccann:
@btakita also reported that Vite optimization breaks Svelte in https://github.com/vitejs/vite/issues/4306

[...]

PR #137, opened 2021-08-10
fix: improve deduplication support 

dominikg:
by adding svelte itself to optimizeDeps.include by default.

See #125 and related discussions. This change should enhance deduplication of svelte itself.

### DIFF
--- a/packages/vite-plugin-svelte/src/utils/options.ts
+++ b/packages/vite-plugin-svelte/src/utils/options.ts
@@ -179,20 +179,24 @@ export function buildExtraViteConfig(
 	options: ResolvedOptions,
 	config: UserConfig
 ): Partial<UserConfig> {
-	const allSvelteImports = [...SVELTE_IMPORTS, ...SVELTE_HMR_IMPORTS];
-
-	// exclude svelte imports from optimization unless explicitly included
-	const excludeFromOptimize = allSvelteImports.filter(
-		(x) => !config.optimizeDeps?.include?.includes(x)
-	);
-
+	// include svelte imports for optimization unless explicitly excluded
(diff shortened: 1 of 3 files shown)


packages/vite-plugin-svelte/src/utils/options.ts
```diff
@@ -179,20 +179,24 @@ export function buildExtraViteConfig(
...
-	const allSvelteImports = [...SVELTE_IMPORTS, ...SVELTE_HMR_IMPORTS];
-
-	// exclude svelte imports from optimization unless explicitly included
-	const excludeFromOptimize = allSvelteImports.filter(
-		(x) => !config.optimizeDeps?.include?.includes(x)
-	);
-
+	// include svelte imports for optimization unless explicitly excluded
+	const include: string[] = [];
+	const exclude: string[] = ['svelte-hmr'];
```
bluwy, line 184: Do we want to add all of `SVELTE_HMR_IMPORTS` here? Or does this already work fine :thinking:

bluwy: Answer from dominik on discord: Vite is able to detect bare package name: https://github.com/vitejs/vite/blob/5c0f0a362f669e3d38ddc51bfdd9d424a7c08feb/packages/vite/src/node/optimizer/scan.ts#L272

bluwy (approved):

packages/vite-plugin-svelte/src/utils/options.ts
dominikg: svelte-hmr in there covers svelte-hmr and all imports starting with svelte-hmr/ so that should work.

PR #145, opened 2021-08-19
feat: auto-exclude svelte dependencies in vite.optimizeDeps

dominikg:
A simple test has been added.

### DIFF
--- a/packages/vite-plugin-svelte/src/utils/dependencies.ts
+++ b/packages/vite-plugin-svelte/src/utils/dependencies.ts
@@ -3,7 +3,7 @@ import path from 'path';
 import fs from 'fs';
 import { createRequire } from 'module';
 
-export function findSvelteDependencies(root: string, cwdFallback = true): SvelteDependency[] {
+export function findRootSvelteDependencies(root: string, cwdFallback = true): SvelteDependency[] {
 	log.debug(`findSvelteDependencies: searching svelte dependencies in ${root}`);
 	const pkgFile = path.join(root, 'package.json');
 	if (!fs.existsSync(pkgFile)) {
@@ -11,10 +11,10 @@ export function findSvelteDependencies(root: string, cwdFallback = true): Svelte
 			const cwd = process.cwd();
 			if (root !== cwd) {
 				log.debug(`no package.json found in vite root ${root}`);
-				return findSvelteDependencies(cwd, false);
+				return findRootSvelteDependencies(cwd, false);
 			}
 		}
-		log.debug(`no package.json found, search failed`);
+		log.warn(`no package.json found, findRootSvelteDependencies failed`);
 		return [];
 	}
 
@@ -26,7 +26,7 @@ export function findSvelteDependencies(root: string, cwdFallback = true): Svelte
 	const deps = [
 		...Object.keys(pkg.dependencies || {}),
 		...Object.keys(pkg.devDependencies || {})
-	].filter((dep) => !excludeFromScan(dep));
+	].filter((dep) => !is_common_without_svelte_field(dep));
 
 	return getSvelteDependencies(deps, root);
 }
@@ -56,7 +56,7 @@ function getSvelteDependencies(
 				dependencyNames = dependencyNames.filter((name) => !path.includes(name));
 			}
 			if (path.length === 3) {
-				log.warn.once(`encountered deep svelte dependency tree ${path.join('>')}`);
+				log.warn.once(`encountered deep svelte dependency tree: ${path.join('>')}`);
 			}
 			result.push(...getSvelteDependencies(dependencyNames, dir, path.concat(pkg.name)));
 		}
@@ -86,7 +86,9 @@ function resolveSvelteDependency(
 				if (!isSvelte(pkg)) {
 					return;
 				}
-				log.warn(`package ${dep} has a "svelte" field but does not export it's package.json`);
+				log.warn.once(
+					`package.json of ${dep} has a "svelte" field but does not include itself in exports field. Please ask package maintainer to update`
+				);
 				return { dir, pkg };
 			}
 			const parent = path.dirname(dir);
@@ -112,7 +114,7 @@ function isSvelte(pkg: Pkg) {
 	return !!pkg.svelte;
 }
 
-const EXCLUDE = [
+const COMMON_DEPENDENCIES_WITHOUT_SVELTE_FIELD = [
 	'@sveltejs/vite-plugin-svelte',
 	'@sveltejs/kit',
 	'autoprefixer',
@@ -120,6 +122,7 @@ const EXCLUDE = [
 	'esbuild',
 	'eslint',
 	'jest',
+	'mdsvex',
 	'postcss',
 	'prettier',
 	'svelte',
@@ -128,7 +131,7 @@ const EXCLUDE = [
 	'typescript',
 	'vite'
 ];
-const EXCLUDE_PREFIX = [
+const COMMON_PREFIXES_WITHOUT_SVELTE_FIELD = [
 	'@postcss-plugins/',
 	'@rollup/',
 	'@sveltejs/adapter-',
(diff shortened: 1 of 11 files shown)


packages/vite-plugin-svelte/src/utils/dependencies.ts
```diff
@@ -0,0 +1,156 @@
+import { log } from './log';
+import path from 'path';
+import fs from 'fs';
+import { createRequire } from 'module';
+
+export function findSvelteDependencies(root: string, cwdFallback = true): SvelteDependency[] {
```
benmccann, line 6: should this be called `findRootSvelteDependencies` to differentiate it from `getSvelteDependencies`?

packages/vite-plugin-svelte/src/utils/dependencies.ts
```diff
@@ -0,0 +1,156 @@
...
+	const pkgFile = path.join(root, 'package.json');
+	if (!fs.existsSync(pkgFile)) {
+		if (cwdFallback) {
+			const cwd = process.cwd();
+			if (root !== cwd) {
+				log.debug(`no package.json found in vite root ${root}`);
+				return findSvelteDependencies(cwd, false);
+			}
+		}
+		log.debug(`no package.json found, search failed`);
```
benmccann, line 17: maybe this should be a `warn` instead of `debug`?

packages/vite-plugin-svelte/src/utils/dependencies.ts
```diff
@@ -0,0 +1,156 @@
...
+	try {
+		const pkgJson = `${dep}/package.json`;
+		const pkg = require(pkgJson);
+		if (!isSvelte(pkg)) {
+			return;
+		}
+		const dir = path.dirname(require.resolve(pkgJson));
+		return { dir, pkg };
+	} catch (e) {
+		log.debug.once(`dependency ${dep} does not export package.json`, e);
```
benmccann, line 78: we should give them some idea of how to address it
```suggestion
		log.debug.once(`package.json of ${dep} does not include itself in exports field. Please ask package maintainer to update`, e);
```

packages/vite-plugin-svelte/src/utils/dependencies.ts
```diff
@@ -0,0 +1,156 @@
...
+	'@rollup/',
+	'@sveltejs/adapter-',
+	'eslint-plugin-',
+	'prettier-plugin-',
+	'postcss-plugin-',
+	'@postcss-plugins/',
+	'@rollup/'
+];
+
+function excludeFromScan(dep: string): boolean {
```
benmccann, line 135: this function could use a comment. it's not clear to me why we're excluding some these

benmccann (commented):
nice!! I'm excited to try this out once the Vite change lands!

packages/vite-plugin-svelte/src/utils/dependencies.ts
dominikg: improved the warning for svelte dependencies instead. I don't think we should put this out here for every dependency that fails. There may be devDependencies that just don't care and the warning would not never go away

packages/vite-plugin-svelte/src/utils/dependencies.ts
dominikg: renamed it and added doc comment. please let me know if thats clear enough

packages/vite-plugin-svelte/src/utils/dependencies.ts
benmccann: `rollup-plugin-svelte` warns for all of them, but gathers the list and then just prints a single warning message at the end: https://github.com/sveltejs/rollup-plugin-svelte/pull/119. That might be a good way to do it so that we let users know there's a package that needs to be updated without spamming them with too many messages

packages/vite-plugin-svelte/src/utils/dependencies.ts
```diff
@@ -0,0 +1,185 @@
...
+	} catch (e) {
+		!silent && log.warn.enabled && log.warn(`failed to parse ${pkgFile}`, e);
+	}
+}
+
+function isSvelte(pkg: Pkg) {
+	return !!pkg.svelte;
+}
+
+const COMMON_DEPENDENCIES_WITHOUT_SVELTE_FIELD = [
```
benmccann, line 117: should we include everything that comes with a new SvelteKit project?
```
  "devDependencies": {
    "@sveltejs/kit": "next",
    "@types/cookie": "^0.4.0",
    "@typescript-eslint/eslint-plugin": "^4.19.0",
    "@typescript-eslint/parser": "^4.19.0",
    "eslint": "^7.22.0",
    "eslint-config-prettier": "^8.1.0",
    "eslint-plugin-svelte3": "^3.2.0",
    "prettier": "~2.2.1",
    "prettier-plugin-svelte": "^2.2.0",
    "svelte": "^3.34.0",
    "svelte-check": "^2.0.0",
    "svelte-preprocess": "^4.0.0",
    "tslib": "^2.0.0",
    "typescript": "^4.0.0"
  },
  "type": "module",
  "dependencies": {
    "@fontsource/fira-mono": "^4.2.2",
    "@lukeed/uuid": "^2.0.0",
    "cookie": "^0.4.1"
  }
```

packages/vite-plugin-svelte/src/utils/dependencies.ts
benmccann: looks good

packages/vite-plugin-svelte/src/utils/dependencies.ts
dominikg: most of it was covered already by the prefix checks, added the remaining ones. thanks!!

packages/vite-plugin-svelte/src/utils/dependencies.ts
```diff
@@ -0,0 +1,190 @@
...
+				log.warn.enabled &&
+					log.warn(
+						`skipping circular svelte dependencies in automated vite optimizeDeps handling`,
+						circular.map((x) => path.concat(x).join('>'))
+					);
+				dependencyNames = dependencyNames.filter((name) => !path.includes(name));
+			}
+			if (path.length === 3) {
+				log.warn.once(`encountered deep svelte dependency tree: ${path.join('>')}`);
+			}
```
bluwy, line 60: Should we be warning when we encounter this scenario? The crawler should be able to continue down the tree, just that it's weird for this to ever happen. I think we can remove these lines, or maybe convert to `debug` if it helps for debugging.

dominikg: Such a deeply nested svelte component package is a sign something is afoot with the project and a warning is in order imho.

I agree that it is weird and tbh i'm hoping this never occurs in the wild. Still i'd like to know if it does and debug would not cut it.

To avoid spam, the warning is logged only once per run and not again if the tree is even deeper. `foo>bar>baz` gets logged, but `foo>bar>baz>qoox` won't.

bluwy: I suppose it's good to know if this ever happens in the wild, but I'm concern if it would annoy the end-user since it may be something not fixable on their end. We might get an issue/pr to remove that line too. Since if it's a warning, the end-user should be able to suppress or fix it (which sometimes isn't possible).

Or we could leave it there and increment the number if someone do report it, so we know there exist an atrocity like that 😆

dominikg:
manually tested it with npm@6 and yarn1 + yarn3 with pnp. and it works :partying_face:

packages/vite-plugin-svelte/src/utils/dependencies.ts
dominikg: made it debug for now. I guess such a deep path would cause other issues and if it cannot be resolved with the package authors we can add back a warning (and an extra option to filter some logs?)

bluwy: Sounds good 👍

PR #157, opened 2021-08-31
wip: enable nested optimization 

dominikg:
- reorganized test dependencies and added cjs/esm deps

Tests are failing for `pnpm test:ci:build` because tests where `simple-svelte` dependency is used , `cjs-only` is added to build output without cjs to esm transform.


bluwy (approved):

packages/e2e-tests/_test_dependencies/cjs-and-esm/index.mjs
```diff
@@ -0,0 +1 @@
+export function cjs_and_esm(){ return 'esm'};
```
benmccann, line 1:
```suggestion
export function cjs_and_esm() { return 'esm' };
```

packages/e2e-tests/_test_dependencies/cjs-and-esm/index.cjs
```diff
@@ -0,0 +1,3 @@
+module.exports = {
+	cjs_and_esm: ()=>'cjs'
```
benmccann, line 2:
```suggestion
	cjs_and_esm: () => 'cjs'
```

dominikg:
Tests fail on windows because somehow the importAnalysis plugin does not replace the cjs import in `Dependency.svelte` of  the `svelte-simple` test dependency.

This works on linux and mac, so my guess is that the `/@fs` prefix on windows is not accounted for correctly in https://github.com/vitejs/vite/blob/c6896e80699bc05791ff1ba6ba665bc683181c28/packages/vite/src/node/plugins/resolve.ts#L482   which leads to an import url that doesn't have the `&es-interop` marker appended and then the import is not replaced with the optimized version.

:/

PR #159, opened 2021-09-01
fix: don't reinclude excluded dependencies in automatic handling

dominikg:
This allows users to prevent certain reincludes that could break vite's optimizer, eg `@types/node`


bluwy (approved):
lgtm if we go down this path

PR #163, opened 2021-09-05
fix: only optimize nested cjs dependencies

### LINKED ISSUE #162: Improve handling for hybrid svelte packages in automatic dependency optimization
### Describe the problem

Some svelte packages are hybrids and could contain dependencies on node-only packages.

These packages then can fail in vite's optimizer with an error like 
`
Failed to resolve entry for package "foo > bar". The package may have incorrect main/module/exports specified in its package.json
`

Currently known hybrids:
* `@roxi/routify` - error caused by `@types/node` (only `types` entry), `cheap-watch`, `rollup-pluginutils` (export esm and used node builtin modules)
* `@carbon/charts-svelte` - error caused by `@carbon/telemetry` (only `bin` entry)

* detect dependencies that would fail in vite's optimizer and don't reinclude them.
* detect dependencies that are not used in svelte components and don't reinclude them

### Alternatives considered

* Improve vite's optimizer to allow deep scanning
* Hardcode list of known hybrids
* Keep the status quo where people have to opt-out of the automatic behavior
  * either by adding the dependencies that fail in vite's optimizer to `optimizeDeps.exclude` in vite config
  * or by adding the declaring package to vite-plugin-svelte's [disableDependencyReinclusion](https://github.com/sveltejs/vite-plugin-svelte/blob/main/docs/config.md#disabledependencyreinclusion) option.
* Change automatic dependency handling to be opt-in in vite-plugin-svelte  

### Importance

would make my life easier

bluwy:
Fixes #162

This would prevent `bin`-only packages, `types`-only packages, and esm packages from pre-bundling. Essentially avoiding the issues noted at https://github.com/sveltejs/vite-plugin-svelte/issues/162#issuecomment-912992931

Also thanks to @dominikg that led to this idea!

### DIFF
--- a/packages/vite-plugin-svelte/src/utils/dependencies.ts
+++ b/packages/vite-plugin-svelte/src/utils/dependencies.ts
@@ -2,7 +2,6 @@ import { log } from './log';
 import path from 'path';
 import fs from 'fs';
 import { createRequire } from 'module';
-import { SVELTE_RESOLVE_MAIN_FIELDS } from './constants';
 
 export function findRootSvelteDependencies(root: string, cwdFallback = true): SvelteDependency[] {
 	log.debug(`findSvelteDependencies: searching svelte dependencies in ${root}`);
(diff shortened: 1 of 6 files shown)


packages/vite-plugin-svelte/src/utils/dependencies.ts
```diff
@@ -174,6 +162,19 @@ function is_common_without_svelte_field(dependency: string): boolean {
 	);
 }
 
+export function needsOptimization(dep: string, localRequire: NodeRequire): boolean {
+	const depData = resolveDependencyData(dep, localRequire);
+	if (!depData) return false;
+	const pkg = depData.pkg;
+	// only optimize if is cjs, using the below as heuristic
```
dominikg, line 169: add a link to #162 so readers have a better chance of figuring out why we do this?

dominikg (approved):

PR #165, opened 2021-09-09
chore: update dependencies; refactor 1 testcase

dominikg:
to be resilient against order changes in an array

somehow after updating the order in that array was changed


bluwy (approved):
Good catch

PR #166, opened 2021-09-11
fix: add svelte lib without components to noExternal

bluwy:
Fixes https://github.com/sveltejs/kit/issues/2397

Note: I can't really add a failing test since the linked `e2e-test-dep-svelte-api-only` package is automatically `noExternal` by default. I can't force it to external via `config.kit.vite.ssr.external` either since our code respects that (not actually testing the bug). For now, I've tested with the issue's repro https://github.com/Nickersoft/urql-repro and it works.

### DIFF
--- a/packages/vite-plugin-svelte/src/utils/dependencies.ts
+++ b/packages/vite-plugin-svelte/src/utils/dependencies.ts
@@ -40,11 +40,13 @@ function getSvelteDependencies(
 	const localRequire = createRequire(`${pkgDir}/package.json`);
 	const resolvedDeps = deps
 		.map((dep) => resolveDependencyData(dep, localRequire))
-		.filter((depData) => !!depData && isSvelte(depData.pkg)) as DependencyData[];
-	// @ts-ignore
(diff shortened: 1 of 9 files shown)


packages/vite-plugin-svelte/src/utils/dependencies.ts
```diff
@@ -102,10 +104,27 @@ function parsePkg(dir: string, silent = false): Pkg | void {
...
+	}
+}
+
+function isSvelteComponentLib(pkg: Pkg) {
 	return !!pkg.svelte;
 }
 
+function isSvelteLib(pkg: Pkg) {
+	return (
+		Object.keys(pkg.dependencies || {}).includes('svelte') ||
```
dominikg, line 123: can be simplified to pkg.dependencies?.svelte

bluwy: That's a lot better

dominikg (approved):

PR #173, opened 2021-09-20
Improve useVitePreprocess

bluwy:
When enabling `useVitePreprocess`, `importsNotUsedAsValues: "preserve"` won't be forced on the entire codebase, only in Svelte files.


dummdidumm:
What is the use case for this? Getting rid of the need to use `svelte-preprocess` when you use TypeScript and nothing else?

bluwy:
Yes, and also when using css preprocessors like scss and stylus. Typescript would be handled with Vite's own esbuild instance so it would replace the need for `svelte-preprocess-esbuild`. Css preprocessors config can be shared via the Vite config too.

There is still use cases for `svelte-preprocess` though, like the recent TypeScript improvements, or it's more advanced preprocessing for style tags.

dominikg (approved):

PR #176, opened 2021-09-22
Remove svelte/ssr resolving

benmccann:
https://github.com/sveltejs/svelte/pull/6743 is a better solution to this problem

### DIFF
--- a/packages/vite-plugin-svelte/src/index.ts
+++ b/packages/vite-plugin-svelte/src/index.ts
@@ -16,7 +16,6 @@ import { VitePluginSvelteCache } from './utils/vite-plugin-svelte-cache';
 import { ensureWatchedFile, setupWatchers } from './utils/watch';
 import { resolveViaPackageJsonSvelte } from './utils/resolve';
 import { addExtraPreprocessors } from './utils/preprocess';
-import { PartialResolvedId } from 'rollup';
 
 export function svelte(inlineOptions?: Partial<Options>): Plugin {
 	if (process.env.DEBUG != null) {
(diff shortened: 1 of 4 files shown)


dominikg:
we need to update peer dependency on svelte to the version that adds the export map and I hope we can keep the current tests instead of removing them. They should still pass with the new solution

Conduitry:
Svelte 3.43.0 has been released with the new export map entry, so we can now try this out for real.

bluwy:
@benmccann We need to update the root `package.json`'s `svelte` version as well since the pnpmfile enforces the same svelte version based on that. So technically the tests are still running `3.42.x`.

We would also want to remove the `ssr.noExternal` handling here:

https://github.com/sveltejs/vite-plugin-svelte/blob/f112d8b4bb029ca98eed500907d3e79308724eb8/packages/vite-plugin-svelte/src/utils/options.ts#L290-L303

but we could do it in a separate PR as this should also work as is. The changes in #166 can also be reverted, and refactored a bit to simplify things.

packages/vite-plugin-svelte/src/index.ts
```diff
@@ -104,25 +101,6 @@ export function svelte(inlineOptions?: Partial<Options>): Plugin {
 				return importee; // query with svelte tag, an id we generated, no need for further analysis
 			}
 
-			if (ssr && importee === 'svelte') {
-				if (!resolvedSvelteSSR) {
-					resolvedSvelteSSR = this.resolve('svelte/ssr', undefined, { skipSelf: true }).then(
```
bluwy, line 109: Since Vite can't resolve `svelte` to the `node` exports, I wonder if we should revert this file change, and swap out `svelte/ssr` to `svelte`, so that we "help" Vite out to resolve it. IIRC `this.resolve`, which uses the `resolver` plugin has support for exports map, so it should work. But I could completely wrong.

If so, I think we should left using `svelte/ssr` at the meantime

bluwy: @benmccann Nevermind, tested this and it doesn't work

PR #177, opened 2021-09-23
chore: update dependencies, pin esbuild to ~0.12 until vite updates to .13

dominikg:
Update dependencies, excluding esbuild 0.13


dominikg:
turns out vite uses incompatible options for esbuild 0.13, so we have to wait until vite made the switch

bluwy (approved):

PR #198, opened 2021-10-17
chore: update dependencies and drop node12 support

dominikg:
Updating kit caused tests running on node12 to fail. We want to follow kit and also require node14.
So this PR updates minimum node version to 14.13.1 and also sets --target node14 for tsup.

Other notable updates: eslint 7 -> 8


dominikg:
we should also check if we can raise tsconfig values for target/module

PR #208, opened 2021-11-05
update deps and node versions on ci

dominikg:
update dependencies
switch ci to node16 and add test for node17


bluwy:
Maybe we can leave out node17 temporarily. Looks like the tests are from the kit-node test, so maybe we can wait for kit to have tests for node17 before we do so. Otherwise I don't have any idea why CI failed too :eyes:

bluwy (approved):

PR #220, opened 2021-12-03
feat: improve compile error reporting

### LINKED ISSUE #218: Better error handling
### Describe the problem

This calls `compile`:

https://github.com/sveltejs/vite-plugin-svelte/blob/031bbda88a05bdd7e9992c6213f51af70a9ccfd3/packages/vite-plugin-svelte/src/utils/compile.ts#L61

But that doesn't give enough details when there's a parse error

It looks like the same issue exists elsewhere - e.g. `esbuild.ts`

I think we need to wrap in a `try` / `catch` and report the `err.frame`

### Alternatives considered

Maybe Svelte itself could give a better error message rather than making the client do it? I've fixed this same issue a whole bunch of places already. But then the client would have less control over display and we'd have to go and update all the clients that already handle the frame

### Importance

would make my life easier

dominikg:
convert compiler error to a RollupError that is displayed to the user in a more friendly way than the current stack pointing into vite-plugin-svelte.

This is a bit verbose and causes messages to be sent multiple times on console for sveltekit (ssr and regular compile) but at least it's showing the correct file and frame.

fixes #218

### DIFF
--- a/packages/vite-plugin-svelte/src/utils/error.ts
+++ b/packages/vite-plugin-svelte/src/utils/error.ts
@@ -2,7 +2,6 @@ import { RollupError } from 'rollup';
 import { Warning } from './options';
 import { buildExtendedLogMessage } from './log';
 import { PartialMessage } from 'esbuild';
-
 /**
  * convert an error thrown by svelte.compile to a RollupError so that vite displays it in a user friendly way
  * @param error
@@ -39,16 +38,24 @@ export function toESBuildError(
 ): PartialMessage {
 	const { filename, frame, start } = error;
 	const partialMessage: PartialMessage = {
-		text: buildExtendedLogMessage(error),
-		detail: frame
+		text: buildExtendedLogMessage(error)
 	};
 	if (start) {
 		partialMessage.location = {
 			line: start.line,
 			column: start.column,
 			file: filename,
-			suggestion: frame
+			lineText: lineFromFrame(start.line, frame) // needed to get a meaningful error message on cli
 		};
 	}
 	return partialMessage;
+}
+
+function lineFromFrame(lineNo: number, frame?: string): string {
+	if (!frame) {
+		return '';
+	}
+	const lines = frame.split('\n');
+	const errorLine = lines.find((line) => line.trimStart().startsWith(`${lineNo}: `));
+	return errorLine ? errorLine.substring(errorLine.indexOf(': ') + 3) : '';
 }
--- a/packages/vite-plugin-svelte/src/utils/error.ts
+++ b/packages/vite-plugin-svelte/src/utils/error.ts
@@ -2,14 +2,13 @@ import { RollupError } from 'rollup';
 import { Warning } from './options';
 import { buildExtendedLogMessage } from './log';
 import { PartialMessage } from 'esbuild';
+
 /**
  * convert an error thrown by svelte.compile to a RollupError so that vite displays it in a user friendly way
- * @param error
+ * @param error a svelte compiler error, which is a mix of Warning and an error
  * @returns {RollupError} the converted error
  */
-export function toRollupError(
-	error: Warning & Error // a svelte compiler error is a mix of Warning and an error
-): RollupError {
+export function toRollupError(error: Warning & Error): RollupError {
 	const { filename, frame, start, code, name } = error;
 	const rollupError: RollupError = {
 		name, // needed otherwise sveltekit coalesce_to_error turns it into a string
@@ -31,12 +30,10 @@ export function toRollupError(
 
 /**
  * convert an error thrown by svelte.compile to an esbuild PartialMessage
- * @param error
+ * @param error a svelte compiler error, which is a mix of Warning and an error
  * @returns {PartialMessage} the converted error
  */
-export function toESBuildError(
-	error: Warning & Error // a svelte compiler error is a mix of Warning and an error
-): PartialMessage {
+export function toESBuildError(error: Warning & Error): PartialMessage {
 	const { filename, frame, start } = error;
 	const partialMessage: PartialMessage = {
 		text: buildExtendedLogMessage(error)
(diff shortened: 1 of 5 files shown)


dominikg:
@benmccann  please try it and let me know if that's good enough.  We may be able to improve this further by additional handling in kit itself.

benmccann:
I think this line needs to be updated as well:

https://github.com/sveltejs/vite-plugin-svelte/blob/aadb92dcc6f110e2e1635387410b43d6b8216f3c/packages/vite-plugin-svelte/src/utils/esbuild.ts#L91

Besides that, lgtm.

> We may be able to improve this further by additional handling in kit itself.

I'm curious what we might do on the kit side?

dominikg:
compile errors are handled for esbuild now too (differently than for vite because it has a different reporting format). 

Should we add similar try/catch to svelte.preprocess ?

bluwy (approved):

benmccann (approved):

dominikg:
i'm having a hard time provoking an error that triggers the esbuild error handling changed in this PR.

dominikg:
finally was able to trigger it thanks to @bluwy .

this is how esbuild prebundling errors will be displayed (cli only)
```shell
> e2e-tests-configfile-esm@1.0.0 dev /home/dominikg/develop/sveltejs/vite-plugin-svelte/packages/e2e-tests/configfile-esm
> vite

 > ../_test_dependencies/svelte-src-export/src/components/Dependency.svelte:3:5: error: [plugin: vite-plugin-svelte:optimize-svelte] /home/dominikg/develop/sveltejs/vite-plugin-svelte/packages/e2e-tests/_test_dependencies/svelte-src-export/src/components/Dependency.svelte:3:5 Unexpected token
     3 │  foo(;
       ╵      ^
   html:/home/dominikg/develop/sveltejs/vite-plugin-svelte/packages/e2e-tests/configfile-esm/src/App.svelte:23:23: note: The plugin "vite-plugin-svelte:optimize-svelte" was triggered by this import
    23 │ import Dependency from 'e2e-test-dep-svelte-src-export/src/components/Dependency.svelte';
       ╵                        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

The link to the file with the error is clickable in intellij idea (and i hope in vscode too)

These esbuild errors should be rare though as the .svelte files are published in a library and hopefully the library maintainers are thoughtful enough to not publish erroneous .svelte files.
Theoretically preprocessor or compiler errors/breaking changes could trigger this for valid .svelte files, but in that case i'd expect a lot more errors everywhere in the project.


vite devserver cli shows this

```shell
3:37:14 PM [vite] Internal server error: /home/dominikg/develop/sveltejs/vite-plugin-svelte/packages/playground/kit-demo-app/src/lib/Counter/index.svelte:14:5 Unexpected token
  Plugin: vite-plugin-svelte
  File: /home/dominikg/develop/sveltejs/vite-plugin-svelte/packages/playground/kit-demo-app/src/lib/Counter/index.svelte
   12 |      return ((n % m) + m) % m;
   13 |    }
   14 |    foo(;
               ^
   15 |  </script>
   16 |  
```
and the overlay looks like this
Screenshot_2021-12-04_15-39-14

bluwy (approved):

packages/vite-plugin-svelte/src/utils/error.ts
```diff
@@ -0,0 +1,100 @@
+import { RollupError } from 'rollup';
+import { Warning } from './options';
+import { buildExtendedLogMessage } from './log';
+import { PartialMessage } from 'esbuild';
+/**
```
bluwy, line 5:
```suggestion

/**
```

A space above

packages/vite-plugin-svelte/src/utils/error.ts
```diff
@@ -0,0 +1,100 @@
...
+ * convert an error thrown by svelte.compile to a RollupError so that vite displays it in a user friendly way
+ * @param error
+ * @returns {RollupError} the converted error
+ */
+export function toRollupError(
+	error: Warning & Error // a svelte compiler error is a mix of Warning and an error
+): RollupError {
```
bluwy, line 12:
```suggestion
 * @param error a svelte compiler error, which is a mix of Warning and an error
 * @returns {RollupError} the converted error
 */
export function toRollupError(error: Warning & Error): RollupError {
```

packages/vite-plugin-svelte/src/utils/error.ts
```diff
@@ -0,0 +1,100 @@
...
+}
+
+/**
+ * convert an error thrown by svelte.compile to an esbuild PartialMessage
+ * @param error
+ * @returns {PartialMessage} the converted error
+ */
+export function toESBuildError(
+	error: Warning & Error // a svelte compiler error is a mix of Warning and an error
+): PartialMessage {
```
bluwy, line 39:
```suggestion
 * @param error a svelte compiler error, which is a mix of Warning and an error
 * @returns {PartialMessage} the converted error
 */
export function toESBuildError(error: Warning & Error): PartialMessage {
```

packages/vite-plugin-svelte/src/utils/error.ts
```diff
@@ -0,0 +1,100 @@
...
+	}
+	return partialMessage;
+}
+
+/**
+ * extract line with number from codeframe
+ *
+ * @param lineNo
+ * @param frame
+ */
```
bluwy, line 60:
```suggestion
/**
 * extract line with number from codeframe
 */
```

packages/vite-plugin-svelte/src/utils/error.ts
```diff
@@ -0,0 +1,100 @@
...
+ * ```
+ *  1 | foo
+ *  2 | bar;
+ *         ^
+ *  3 | baz
+ * ```
+ * @see https://github.com/vitejs/vite/blob/96591bf9989529de839ba89958755eafe4c445ae/packages/vite/src/client/overlay.ts#L116
+ *
+ * @param frame
+ */
```
bluwy, line 91:
```suggestion
 */
```

bluwy (commented):
Some format nitpicks

PR #223, opened 2021-12-08
refactor: use new server.restart api and update vite peer to ^2.7.0 

dominikg:
instead of fake change event on config file

### DIFF
--- a/.changeset/smooth-rings-run.md
+++ b/.changeset/smooth-rings-run.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/vite-plugin-svelte': major
+---
+
+update vite peerDependency to ^2.7.0 and refactor server restart on change of svelte.config.js
--- a/packages/vite-plugin-svelte/package.json
+++ b/packages/vite-plugin-svelte/package.json
@@ -55,7 +55,7 @@
   "peerDependencies": {
     "diff-match-patch": "^1.0.5",
     "svelte": "^3.44.0",
-    "vite": "^2.6.0"
+    "vite": "^2.7.0"
   },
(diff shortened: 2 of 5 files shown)


dominikg:
unfortunately this doesn't work for sveltekit out of the box (see packages/e2e-test/kit-node )
On restart, index.html is not found/loaded in the browser. needs investigation

dominikg:
The reason it doesn't work for kit is that kit does extra work when it initializes the devserver instance the first time which is not repeated when we call server.restart()

Pointers:
https://github.com/sveltejs/kit/blob/ae7f7904c4f3c111dbff1ad1209ad92fb6f057cf/packages/kit/src/cli.js#L94
https://github.com/sveltejs/kit/blob/master/packages/kit/src/core/dev/index.js

Esp. removing middlewares would have to be done again
https://github.com/sveltejs/kit/blob/ae7f7904c4f3c111dbff1ad1209ad92fb6f057cf/packages/kit/src/core/dev/index.js#L555
[9:22 PM] related: https://github.com/vitejs/vite/pull/4640 and https://github.com/vitejs/vite/issues/5720

this could be achieved by using a special "vite-plugin-kit" that implements configResolved, configureServer and buildStart hooks to update vite instead of doing it once. 

If this is wanted for kit, it should be implemented in a PR in kit repo, once that has been merged we can remove the "not for kit" constraint in here.

bluwy (approved):
LGTM. Re sveltekit, looks like the code is a fine compromise for now until we find a cleaner way.

dominikg:
don't merge/release until kit repo has been updated to vite 2.7

benmccann (approved):

PR #225, opened 2021-12-13
fix: ignore protocol when resolve svelte field

bluwy:
Fixes the CI issue in https://github.com/sveltejs/kit/pull/3034 where the `node:` protocol was messing with resolving the `svelte` field of package.json. Adding `node-fetch`, Vite's scanner seems to resolve its nested imports using Vite plugin's `resolveId`, and because we're using `enforce: "pre"`, we encounter this scenario before Vite's default resolver, which would've handled this.

I'm also not sure why it's resolving the nested imports, but at the meantime this addition looks fine as well.

### DIFF
--- a/packages/vite-plugin-svelte/src/utils/resolve.ts
+++ b/packages/vite-plugin-svelte/src/utils/resolve.ts
@@ -14,7 +14,13 @@ export function resolveViaPackageJsonSvelte(importee: string, importer?: string)
 }
 
 function isBareImport(importee: string): boolean {
-	if (!importee || importee[0] === '.' || importee[0] === '\0' || path.isAbsolute(importee)) {
+	if (
+		!importee ||
+		importee[0] === '.' ||
+		importee[0] === '\0' ||
+		importee.includes(':') ||
+		path.isAbsolute(importee)
+	) {
 		return false;
 	}
 	const parts = importee.split('/');
--- a/.changeset/clean-rice-fly.md
+++ b/.changeset/clean-rice-fly.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/vite-plugin-svelte': patch
+---
+
+Ignore import protocols like `node:` when resolving the `svelte` field in package.json


packages/vite-plugin-svelte/src/utils/resolve.ts
```diff
@@ -14,7 +14,13 @@ export function resolveViaPackageJsonSvelte(importee: string, importer?: string)
 }
 
 function isBareImport(importee: string): boolean {
-	if (!importee || importee[0] === '.' || importee[0] === '\0' || path.isAbsolute(importee)) {
+	if (
+		!importee ||
+		importee[0] === '.' ||
+		importee[0] === '\0' ||
+		importee.includes(':') ||
```
benmccann, line 21: Should this check if it starts with `node:`? I'm guessing we're unlikely to run into a `:` elsewhere, but I didn't know if there are other prefixes that you might see besides `node`?
```suggestion
		importee.startsWith('node:') ||
```

benmccann (approved):

packages/vite-plugin-svelte/src/utils/resolve.ts
Conduitry, line 21: Should we just do `.startsWith('node:')`? What else is this trying to catch?

bluwy: According to [node](https://nodejs.org/api/esm.html#urls), this would also catch `file:` and `data:`. And sometimes virtual imports uses `:` too, so we can catch it here.

bluwy: I replied at https://github.com/sveltejs/vite-plugin-svelte/pull/225#discussion_r768003988

Conduitry: `:` is not allowed in paths in Windows, but it is on Linux (and presumably other UNIXes), which is making me concerned about handling this broadly like this. Would `/^[a-z]+:/.test(importee)` work?

bluwy: I think it should still be fine since we want to make sure `importee` is a node_modules package. So it has to be something like `foobar` or `@foo/bar` for this function to return true. And package names can't contain `:`. Unless I'm missing something 🤔

Conduitry: Ah, I see, yep. Since this is only concerned with whether this is a bare import of the top level of a `node_modules` package, then even something like `foo/bar:baz` being caught by this would be fine, since it would be caught later on anyway.

Conduitry (approved):

packages/vite-plugin-svelte/src/utils/resolve.ts
benmccann: ok. sounds good

PR #235, opened 2021-12-23
chore: non-major dependency updates

dominikg:


bluwy (commented):
I think we'd want to update `pnpm.overrides.vite` as well to "^2.7.6". That likely caused the stale lockfile (not sure why pnpm does that still)

bluwy (approved):

PR #237, opened 2021-12-28
feat: auto-restart SvelteKit when Svelte config changed

### LINKED ISSUE #210: Automatic Restart After Config Change
### Describe the problem

When changing `svelte.config.js` during `dev` runtime, one gets an error message:
```
[vite-plugin-svelte] Svelte config change detected, restart your dev process to apply the changes.
```

Create an option in `svelte.config.js` that enables automatic restart.

This is not a new idea. For example, C# (dotnet) support this. See "rude edit" [here](https://devblogs.microsoft.com/dotnet/update-on-net-hot-reload-progress-and-visual-studio-2022-highlights/#improved-user-experience-in-visual-studio-2022).

### Alternatives considered

Maybe this should be added to `vite-plugin-svelte` instead?

### Importance

nice to have

### Additional Information

_No response_

bluwy:
Closes #210

Tested locally and it seems that it's working fine for SvelteKit on latest version. It probably won't work when someone updates vite-plugin-svelte and not sveltekit, but perhaps it's fine since we're in pre mode.

### DIFF
--- a/packages/vite-plugin-svelte/src/utils/options.ts
+++ b/packages/vite-plugin-svelte/src/utils/options.ts
@@ -99,7 +99,7 @@ export function resolveOptions(
 	viteConfig: ResolvedConfig
 ): ResolvedOptions {
 	const defaultOptions: Partial<Options> = {
-		hot: viteConfig.isProduction ? false : { injectCss: !preResolveOptions.emitCss },
+		hot: viteConfig.isProduction ? false : { injectCss: preResolveOptions.emitCss },
 		compilerOptions: {
 			css: !preResolveOptions.emitCss,
 			dev: !viteConfig.isProduction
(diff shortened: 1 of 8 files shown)


packages/vite-plugin-svelte/src/utils/options.ts
```diff
@@ -101,7 +99,7 @@ export function resolveOptions(
 	viteConfig: ResolvedConfig
 ): ResolvedOptions {
 	const defaultOptions: Partial<Options> = {
-		hot: viteConfig.isProduction ? false : { injectCss: preResolveOptions.emitCss },
+		hot: viteConfig.isProduction ? false : { injectCss: !preResolveOptions.emitCss },
```
bluwy, line 102: My bad. This should have `!` based on the previous code. Got lost after the refactor. Otherwise the `enforceOptionsForHmr()` function would make a warning on startup everytime.

bluwy: Removed this change, and moved to #238

dominikg (approved):

PR #242, opened 2022-01-01
chore: update major dependencies

bluwy:
Including minor dependencies.

`execa` and `node-fetch` aren't updated since the latest major version are ESM only, which wouldn't work in Jest.

`sirv` is removed, since it wasn't used anywhere.


bluwy:
Blocked by https://github.com/sveltejs/kit/pull/3169 and https://github.com/sveltejs/kit/issues/3159. I've downgraded SvelteKit to next.210 to fix the first issue, but now the CI fails on the second issue. I think we can fix them first, then continue on this PR.

bluwy:
Test failed twice:

1. SvelteKit hydrates faster than expected, fails the `SERVER_LOADED` check (It's already `CLIENT_LOADED`)
2. Two continuous server `console.log` apparently has the chance of being a single string from the string array `e2eServer.logs.server.out`

These are strange but this PR should be safe now.

dominikg (approved):

Issue #222, opened 2021-12-08
Cleanup prebundleSvelteLibraries


bluwy:
### Describe the problem

Fix the remaining issues for `prebundleSvelteLibraries`, which should be the correct way forward for Vite prebundling of Svelte libraries. Namely:

- [x] Automatic re-prebundle when svelte config change.
- [x] Allow deep importing Svelte components.
- [x] Allow importing libraries with a Svelte component entrypoint.
- [x] Make sure Vite 2.9 prebundling changes works

Some fixes may need to be in Vite core. Collecting these issue here so I don't forget, or if someone likes to tackle them.

### Describe the proposed solution

Goal is to fix them by year end.

### Alternatives considered

Not fix them by year end

### Importance

would make my life easier

bluwy:
Update: no2 and no3 are being dealt at https://github.com/vitejs/vite/pull/6801

josdejong:
Short question: what will be the default behavior after this is finished? Will that be `prebundleSvelteLibraries: true`?

Background: I'm creating a sveltekit library ([see here](https://github.com/josdejong/svelte-jsoneditor/)), but in order to be able to use it right now, every client has to configure `kit.vit.optimizeDeps.include: [...list with commonjs deps...]`. I noticed that this is not needed anymore when setting `experimental.prebundleSvelteLibraries: true`, which is great! So I hope this will become the default in the end: that will make using my library a breeze (just install and use, no extra config needed).

bluwy:
Hi. Ideally this will be the new default behaviour, but I would put it in a short experimental phase first just in case there are edge cases.

> I'm creating a sveltekit library ([see here](https://github.com/josdejong/svelte-jsoneditor/)), but in order to be able to use it right now, every client has to configure `kit.vit.optimizeDeps.include: [...list with commonjs deps...]`

This should be handled by default by `vite-plugin-svelte` already ([nested/transitive dependency inclusion](https://github.com/sveltejs/vite-plugin-svelte/pull/157)), so it's weird that that's needed for your library. I think the `optimizeDeps.include` can be removed now and it will work, but feel free to open an issue if it doesn't.

josdejong:
Awesome!

Good to know that this is supposed to "just" work. I'll double check whether I'm using all the latest versions of svelte/kit/vite and verify whether `optimizeDeps.include` is still needed, if so I'll open a new issue, thanks Bjorn.

PR #245, opened 2022-01-02
feat: re-prebundle when config changed

bluwy:
Resolve `Automatic re-prebundle when svelte config change.` task of #222.

I think this works nicely based on my brief tests. The `preprocessor` field should also be taken care of by converting them into `.toString()`, which is good enough.

### DIFF
--- a/packages/vite-plugin-svelte/src/utils/watch.ts
+++ b/packages/vite-plugin-svelte/src/utils/watch.ts
@@ -54,7 +54,7 @@ export function setupWatchers(
 			});
 		} else {
 			log.info(`svelte config changed: restarting vite server. - file: ${filename}`);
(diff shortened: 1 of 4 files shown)


packages/vite-plugin-svelte/src/utils/watch.ts
```diff
@@ -54,7 +54,7 @@ export function setupWatchers(
 			});
 		} else {
 			log.info(`svelte config changed: restarting vite server. - file: ${filename}`);
-			server.restart(!!options.experimental?.prebundleSvelteLibraries);
+			server.restart();
```
bluwy, line 57: This was removed as `buildStart` will now conditionally determine if prebundling (or force optimize) is needed. Instead of always prebundle on restart.

dominikg (approved):

PR #247, opened 2022-01-06
fix: use the resolved vite root

ElMassimo:
### Description 📖 

This pull request fixes a [compatibility issue] between `vite-plugin-svelte` and backend integrations like [Vite Ruby], by using the `root` provided by Vite in the `configResolved` hook.

### Background 📜 

`vite-plugin-svelte` needs to infer the `root` before `configResolved` in order to pre-scan svelte dependencies, which was last changed in #115.

Vite plugins can override options using the `config` hook. In particular, [Vite Ruby](https://github.com/ElMassimo/vite_ruby) changes the `root` option based on user configuration, which is typically a subfolder (instead of the project root, `process.cwd()`).

`vite-plugin-svelte` does not use the resolved value for `root`, which can cause Vite and `vite-plugin-svelte` to use different values of `root` to resolve paths.

### The Bug 🐞 

Since the `root` is different, the [check for file existence](https://github.com/sveltejs/vite-plugin-svelte/blob/main/packages/vite-plugin-svelte/src/utils/id.ts#L70) when creating a CSS virtual module no longer matches the expected result, preventing CSS virtual modules from [being served][compatibility issue].

### Proposed Fix  🔨 

Using the `root` that Vite provides in the `configResolved` hook fixes this and other similar problems.

### Notes ✏️ 

Since `vite-plugin-svelte` uses `enforce: 'pre'`, it's not viable for other plugins to workaround this issue without this patch.

Using `enforce: 'pre'` in other plugins would make the outcome dependent on the order in which the plugins are applied in the user config, which would be error-prone and result in a bad user experience.

[Vite Ruby]: https://github.com/ElMassimo/vite_ruby
[compatibility issue]: https://github.com/ElMassimo/vite_ruby/issues/171

### DIFF
--- a/.changeset/popular-tigers-camp.md
+++ b/.changeset/popular-tigers-camp.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/vite-plugin-svelte': patch
+---
+
+use the resolved vite root to support backend integrations
--- a/packages/vite-plugin-svelte/src/utils/options.ts
+++ b/packages/vite-plugin-svelte/src/utils/options.ts
@@ -112,6 +112,7 @@ export function resolveOptions(
 			...defaultOptions.compilerOptions,
 			...preResolveOptions.compilerOptions
 		},
+		root: viteConfig.root,
 		isProduction: viteConfig.isProduction
 	};
 	addExtraPreprocessors(merged, viteConfig);


bluwy:
Hey 👋 Thanks for bring this up. This looks like a valid issue, since we have always worked under the assumption that `root` (and other usual user configs) are never changed by a plugin. Good to know that there's a case like this out there.

This brings another (minor) issue though. We use the `root` to look for `svelte.config.js`, and loading the Svelte config has to happen in the `config` hook since it would return additional Vite config. With this PR, the `svelte.config.js` would be loaded from the old `root` instead.

This likely isn't a big issue, but just noting here for future reference.

---

Re the change, this looks fine to me. But on our part we might have to re-think how we should load Svelte configs.

dominikg:
Thank you for this high quality PR. :heart: 

Regarding `enforce: pre`, this has been required for vite-plugin-svelte to be able to resolve svelte correctly. We may be able to either split vite-plugin-svelte into multiple plugins where only the resolve part has 'pre' or check if changes in vite since then made it possible to do it differently.

But in the interim i'm going to merge this, as it doesn't change the behavior if root is unchanged and improves the behavior if it is. 

Although i think plugins should not modify root from their config hook. This value is fundamental and should be stable for all plugins regardless or ordering. Users can provide a custom root from their config file or via cli. Custom frameworks or integrations could document how to set that up or provide a wrapper around `defineConfig` :thinking: 

Imagine a situation where 2 plugins want to modify root to different locations. Which one "wins"?

dominikg (approved):

buhrmi:
Just updated to 1.0.0-next.34, works perfectly. Thanks guys

PR #254, opened 2022-01-19
fix: replace relative-require with our own resolveDependencyData

### LINKED ISSUE #244: Warning 'The following packages did not export their `package.json` file...' in case of non-svelte packages
### Describe the problem

I'm always getting a warning message  from `vite:resolve` when tslib is resolved:
```[vite-plugin-svelte] The following packages did not export their package.json file so we could not check the "svelte" field. If you had difficulties importing svelte components from a package, then please contact the author and ask them to export the package.json file.```
```- tslib``` 
(see [index.ts:205](https://github.com/sveltejs/vite-plugin-svelte/blob/main/packages/vite-plugin-svelte/src/index.ts#L205))
The message is caused when somewhere in `src/utils/resolve.ts:resolveViaPackageJsonSvelte(..)` the error `ERR_PACKAGE_PATH_NOT_EXPORTED` occurs.
Because `tslib` Is a third-party lib that does not use Svelte, it does not make sense to check the "svelte" field.

It would be nice to have an option to exclude some libs from being checked for the "svelte" field.

### Alternatives considered

None

### Importance

nice to have

dominikg:
to avoid errors when package.json isn't exported

this could resolve #244  and also should get rid of errors that package.json isn't exported in general as resolveDependencyData falls back to a filesystem read - and even if it isn't exported, npm packages do always contain their package.json . 

Note: before merging we should check performance and make sure it works the way we expect it to.

Some libraries (eg carbon-components-svelte) ship both compiled and uncompiled code, svelte field must take precedence in this case.

( related: https://github.com/carbon-design-system/carbon-components-svelte/pull/991 )


dominikg:
updated to latest vite too as for some reason pnpm i left it in an inconsistent state for me (maybe because of the override that wasn't updated by renovate?)

dominikg:
note to future self: cache results for resolved svelte fields

bluwy (approved):
LGTM. The implementation looks simpler than I expected, which is great. I haven't tested this locally though but it looks like it'll work.

bluwy (approved):
Nice. The caching looks good.

PR #262, opened 2022-01-29
fix: change cjs detection heuristic to take exports map into account

dominikg:
"exports" in package.json isn't only for esm and if it exists it takes precedence over "main". 
We can't realistically parse it to find the entry ourselves, so delegate that to require.resolve

Questions: is that the right approach? If so, should we simplify it and do it regardless of presence of main/exports?

cc @benmccann

### DIFF
--- a/packages/vite-plugin-svelte/src/utils/dependencies.ts
+++ b/packages/vite-plugin-svelte/src/utils/dependencies.ts
@@ -187,12 +187,23 @@ export function needsOptimization(dep: string, localRequire: NodeRequire): boole
 	const pkg = depData.pkg;
 	// only optimize if is cjs, using the below as heuristic
 	// see https://github.com/sveltejs/vite-plugin-svelte/issues/162
-	const isCjs = pkg.main && !pkg.module && !pkg.exports;
-	if (!isCjs) return false;
-	// ensure entry is js so vite can prebundle it
-	// see https://github.com/sveltejs/vite-plugin-svelte/issues/233
-	const entryExt = path.extname(pkg.main);
-	return !entryExt || entryExt === '.js' || entryExt === '.cjs';
+	if (pkg.type !== 'module' && !pkg.module) {
+		let entry;
+		// if exports is defined it takes precedence over main!
+		if (pkg.exports) {
+			// exports map syntax is too complex to try and find the correct one, let node do it's thing
+			entry = localRequire.resolve(dep);
+		} else if (pkg.main) {
+			entry = pkg.main;
+		}
+		if (entry) {
+			const ext = path.extname(entry);
+			// ensure entry is js so vite can prebundle it
+			// see https://github.com/sveltejs/vite-plugin-svelte/issues/233
+			return !ext || ext === '.js' || ext === '.cjs';
+		}
+	}
+	return false;
 }
 
 interface DependencyData {
--- a/packages/e2e-tests/_test_dependencies/cjs-and-esm-exports/index.cjs
+++ b/packages/e2e-tests/_test_dependencies/cjs-and-esm-exports/index.cjs
@@ -1,0 +1,3 @@
+module.exports = {
+	cjs_and_esm_exported: () => 'cjs-via-exports'
+}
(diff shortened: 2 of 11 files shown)


packages/vite-plugin-svelte/src/utils/dependencies.ts
```diff
@@ -187,12 +187,23 @@ export function needsOptimization(dep: string, localRequire: NodeRequire): boole
 	const pkg = depData.pkg;
 	// only optimize if is cjs, using the below as heuristic
 	// see https://github.com/sveltejs/vite-plugin-svelte/issues/162
-	const isCjs = pkg.main && !pkg.module && !pkg.exports;
-	if (!isCjs) return false;
-	// ensure entry is js so vite can prebundle it
-	// see https://github.com/sveltejs/vite-plugin-svelte/issues/233
-	const entryExt = path.extname(pkg.main);
-	return !entryExt || entryExt === '.js' || entryExt === '.cjs';
+	if (pkg.type !== 'module' && !pkg.module) {
```
dominikg, line 190: the pkg.module check here has been kept from the previous heuristic. It assumes that vite will resolve via that module field.

benmccann: also, @bluwy pointed out to me the other day that even if `pkg.type === 'module'` you can still have a `.cjs` in the `main` and so we don't know that it's ESM. I've never seen that in the wild, but it's technically possible

packages/vite-plugin-svelte/src/utils/dependencies.ts
```diff
@@ -187,12 +187,23 @@ export function needsOptimization(dep: string, localRequire: NodeRequire): boole
...
+		let entry;
+		// if exports is defined it takes precedence over main!
+		if (pkg.exports) {
+			// exports map syntax is too complex to try and find the correct one, let node do it's thing
+			entry = localRequire.resolve(dep);
```
bluwy, line 195: I think for `pkg.exports` we can safely assume that most packages export an ESM variant. `exports` for CJS-only sounds rare and probably not worth the effort to check for these.

I think the previous heuristic we have is fine, with an extra change on the final extension check like:

```js
return !entryExt || (entryExt === '.js' && pkg.type !== 'module') || entryExt === '.cjs';
```

bluwy: Another risk with using resolve might be that it's resolved based on node conditions, but during prebundling, it'll be bundled for the browser. This won't likely be a problem, but just something I thought of.

packages/vite-plugin-svelte/src/utils/dependencies.ts
dominikg: > If `exports` is defined, I think that will take precedence over `pkg.module` though?

module is only ever used by bundlers, so it would be up to them. The presence is a strong hint that it supports esm when bundled. I think vite honors module but have to check

packages/vite-plugin-svelte/src/utils/dependencies.ts
dominikg: that is true. Wonder if we can set the browser condition

packages/vite-plugin-svelte/src/utils/dependencies.ts
benmccann: Vite does use `module`. I suppose it's fair to say that if `module` is defined then it supports ESM

packages/vite-plugin-svelte/src/utils/dependencies.ts
dominikg: unfortunately it is not possible for us to set custom user conditions apart from passing a flag on the cli :see_no_evil:  https://nodejs.org/api/packages.html#resolving-user-conditions

dominikg:
So we're going to put this on hold. There is no perfect way for us to detect cjs and the current heuristic has worked well. Until actual issues are being reported there's no point it risking stability and once svelte dependency prebundling is default, it should become a non-issue anyways.

unsolved problems:
- even with require, we might not find the right entry as vite could resolve differently than node itself
- we'd have to find the package.json closest to entrypoint to check `"type": "module"` instead of relying on package itself

leaving notes: 
- this hole thing started from https://github.com/microsoft/monaco-editor/pull/2921 which was triggered by a discussion in sveltekit

PR #276, opened 2022-02-17
Extend docs for `useVitePreprocess`

benmccann:
Adds a bit more detail. It wasn't exactly clear to me what the option did at first. It looks very cool now that I've learned a bit more about it!


bluwy (approved):
Nice!

dominikg (approved):

PR #280, opened 2022-02-26
fix: remove svelte css transform

bluwy:
Fixes https://github.com/sveltejs/kit/issues/4125

I can't seem to transfer it over to this repo. GitHub seems to have issues.

### DIFF
--- a/.changeset/slow-buses-tap.md
+++ b/.changeset/slow-buses-tap.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/vite-plugin-svelte': patch
+---
+
+Remove transforming svelte css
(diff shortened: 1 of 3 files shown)


dominikg:
i tried to do this locally yesterday but had some issues testing that it works as expected. 

Ideally we'd add an e2e test that adds 3 transform plugins (pre, regular, post) that do both css and js transforms to a test svelte component and see that they all actually do work. 
In addition to a regular vite+svelte setup a similar test for sveltekit may be needed too. 

In a broader sense i wonder if there is a way to build a testsuite to validate that vite plugins behave per spec.

bluwy:
Added vite plugin tests for the `hmr` and `kit-node` e2e tests. All seems to be working fine now.

> In a broader sense i wonder if there is a way to build a testsuite to validate that vite plugins behave per spec.

I think VIte has the potential for testing those explicitly in the [plugin container spec file](https://github.com/vitejs/vite/blob/main/packages/vite/src/node/server/__tests__/pluginContainer.spec.ts), but it's not a lot of coverage right now. The e2e tests sort-of close the gap over time.

dominikg (approved):

Issue #283, opened 2022-02-27
npm run build retrieves `The argument 'filename' must be a file URL object, file URL string, or absolute path string`


silvestrevivo:
### Describe the bug

I've created a npm package that in development it works OK but when I run `npm run build` it retrieves the next error. 

```
[vite-plugin-svelte] The argument 'filename' must be a file URL object, file URL string, or absolute path string. Received '\x00__vite-browser-external?commonjs-proxy'
> The argument 'filename' must be a file URL object, file URL string, or absolute path string. Received '\x00__vite-browser-external?commonjs-proxy'
```

### Reproduction

This is the repo where this is happening. 

https://github.com/silvestrevivo/solana-svelte-counter

I think it would be difficult to reproduce since you need to deploy the smart contract to run the project. **With the SvelteKit version `251` this bug doesn't occur.** As soon as you upgrade SvelteKit, the error happens.  

### Logs

```shell
[vite-plugin-svelte] The argument 'filename' must be a file URL object, file URL string, or absolute path string. Received '\x00__vite-browser-external?commonjs-proxy'
> The argument 'filename' must be a file URL object, file URL string, or absolute path string. Received '\x00__vite-browser-external?commonjs-proxy'
    at new NodeError (node:internal/errors:371:5)
    at createRequire (node:internal/modules/cjs/loader:1210:13)
    at resolveViaPackageJsonSvelte (file:///Users/silvestrevivo/werk/2021/solana/solana-svelte-counter/app/node_modules/@sveltejs/vite-plugin-svelte/dist/index.js:1427:26)
    at Object.resolveId (file:///Users/silvestrevivo/werk/2021/solana/solana-svelte-counter/app/node_modules/@sveltejs/vite-plugin-svelte/dist/index.js:1585:24)
    at /Users/silvestrevivo/werk/2021/solana/solana-svelte-counter/app/node_modules/rollup/dist/shared/rollup.js:22795:37
```

### System Info

```shell
.
```

### Severity

annoyance

ashwinvasudevan:
try, catch has been removed around `resolveViaPackageJsonSvelte` fn. Adding try catch seems to solve the error. Downgrade to `@sveltejs/vite-plugin-svelte@1.0.0-next.37` to solve the issue for now

silvestrevivo:
[...]

PR #286, opened 2022-02-28
fix: prevent errors in resolveViaPackageJsonSvelte breaking vite resolve

dominikg:
see #283


dominikg:
should we improve resolveViaPackageJsonSvelte to gracefully handle artificial ids that don't work?

bluwy (approved):

Issue #281, opened 2022-02-26
Error including `bytemd` dependency


benmccann:
### Describe the bug

Someone  on Discord reported that they can't get `bytemd` to work. It looks to me like `vite-plugin-svelte` is not detecting that `bytemd`'s `lodash.debounce` dependency needs to be prebundled. There are also issues reported in the `bytemd` repo and [on Discord](https://discord.com/channels/457912077277855764/946968773045411880/946968774047830087) where the error occurs on other dependencies

A couple things I noticed that are off about the packaging:
- `bytemd` uses `.esm.js` and `.cjs.js` extensions instead of `.mjs` and `.cjs`
- `lodash.debounce` has no `main` or any other entry point specified in the `package.json`

Even when I manually edit those I still get the same error.

### Reproduction

https://github.com/benmccann/sveltekit-bytemd-bug-reproduction

### Logs

```shell
8:07:24 PM [vite] Error when evaluating SSR module /node_modules/lodash.debounce/index.js?v=dc88d06f:
ReferenceError: module is not defined
    at /node_modules/lodash.debounce/index.js?v=dc88d06f:377:1
```

### System Info

```shell
reproduction provided
```

### Severity

The top two bugs in the `bytemd` repo are about how the project doesn't work with SvelteKit: https://github.com/bytedance/bytemd/issues/96 and https://github.com/bytedance/bytemd/issues/139

bluwy:
My bad. I assumed this was a pre-bundling bug and only fixed that, looks like the error is coming from loading the module in SSR 🤔

bluwy:
TL;DR: `ssr.noExternal` on a dep considers the dep to be source code (Everything is ESM including its transitive deps, like pre-bundling)

---

Wow this seems like a huge deal than I expected. Through some testing, I realized the issue is that `ssrLoadModule` is called on `lodash.debounce`, but that doesn't work because `ssrLoadModule` works for ESM code only. Ideally Vite expects this to never happen in the first place.

For our case, it happens because `ssrLoadModule` is [called by ssrImport](https://github.com/vitejs/vite/blob/main/packages/vite/src/node/ssr/ssrModuleLoader.ts#L143), which happens when you do `import debounce from "lodash.debounce"` in your code. And in this case, `bytemd`'s is [doing so](https://github.com/bytedance/bytemd/blob/1981d8ab1a4979e9ac43440a5029e7b63191ded6/packages/bytemd/src/helpers.ts#L4).

[...]

PR #289, opened 2022-02-28
fix: externalize transitive cjs only deps during dev ssr

dominikg:
see #281 

i'm still getting an error though when adding `bytemd` to a sveltekit project 

this is rendered in the html output of localhost:3000 after a brief flash of an unstyled editor.
```
500
The requested module '/node_modules/.pnpm/codemirror-ssr@0.0.6/node_modules/codemirror-ssr/mode/gfm/gfm.js?v=9e01fcfd' does not provide an export named 'default'
SyntaxError: The requested module '/node_modules/.pnpm/codemirror-ssr@0.0.6/node_modules/codemirror-ssr/mode/gfm/gfm.js?v=9e01fcfd' does not provide an export named 'default'
```

### DIFF
--- a/packages/vite-plugin-svelte/src/utils/options.ts
+++ b/packages/vite-plugin-svelte/src/utils/options.ts
@@ -318,9 +318,10 @@ function buildSSROptionsForSvelte(
 
 	if (options.isServe) {
 		// during dev, we have to externalize transitive dependencies, see https://github.com/sveltejs/vite-plugin-svelte/issues/281
-		const transitiveDepsOfSvelteLibs = [
-			...new Set(svelteDeps.flatMap((dep) => Object.keys(dep.pkg.dependencies || {})))
-		].filter(
+		// @ts-expect-error ssr still flagged in vite
+		ssr.external = Array.from(
+			new Set(svelteDeps.flatMap((dep) => Object.keys(dep.pkg.dependencies || {})))
+		).filter(
 			(dep) =>
 				!ssr.noExternal.includes(dep) &&
 				// @ts-expect-error ssr still flagged in vite
@@ -328,8 +329,6 @@ function buildSSROptionsForSvelte(
 				// @ts-expect-error ssr still flagged in vite
 				!config.ssr?.external?.includes(dep)
 		);
-		// @ts-expect-error ssr still flagged in vite
-		ssr.external = transitiveDepsOfSvelteLibs;
 	}
 
 	return ssr;
--- a/packages/vite-plugin-svelte/src/utils/options.ts
+++ b/packages/vite-plugin-svelte/src/utils/options.ts
@@ -184,8 +184,7 @@ function resolveViteRoot(viteConfig: UserConfig): string | undefined {
 
 export function buildExtraViteConfig(
 	options: PreResolvedOptions,
-	config: UserConfig,
-	configEnv: ConfigEnv
+	config: UserConfig
 ): Partial<UserConfig> {
 	// extra handling for svelte dependencies in the project
 	const svelteDeps = findRootSvelteDependencies(options.root);
@@ -200,7 +199,7 @@ export function buildExtraViteConfig(
 		// knownJsSrcExtensions: options.extensions
 	};
 
-	if (configEnv.command === 'serve') {
+	if (options.isServe) {
 		extraViteConfig.optimizeDeps = buildOptimizeDepsForSvelte(
 			svelteDeps,
 			options,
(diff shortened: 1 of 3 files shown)


benmccann:
I was thinking that this is probably something that should be fixed purely in Vite. Vite is already including the packages in `externals`, but it's just that they're listed as `parent > child` so either we should make it understand that syntax or stop generating that syntax in that code path

dominikg:
totally fine with fixing it in vite :+1:   anyone got an idea why it's failing for codemirror-ssr though?

benmccann:
I didn't look at that one yet, but codemirror is probably the hardest library I've ever encountered to get working in Vite, so I'm not terribly surprised or worried about that one. I shared with the bytemd how we got the main codemirror package working for svelte.dev

packages/vite-plugin-svelte/src/utils/options.ts
```diff
@@ -211,6 +211,24 @@ export function buildExtraViteConfig(
...
+				// @ts-expect-error ssr still flagge in vite
+				extraViteConfig.ssr.external = [];
+			}
+			// @ts-expect-error ssr still flagge in vite
+			extraViteConfig.ssr.external.push(
+				// @ts-expect-error ssr still flagge in vite
+				...optimizedTransitiveDeps.filter((x) => !config.ssr?.noExternal?.includes(x))
+			);
+		}
+	}
```
bluwy, line 231: I think we could move this into `buildSSROptionsForSvelte()` and build the list of externals from scratch, since if `prebundleSvelteLibraries` is turned off, there won't be `parent > child` syntax to handle.

bluwy (commented):
Ben made https://github.com/vitejs/vite/pull/7154 which should also fix this, but I think we should also merge this one as when someone turns on `prebundleSvelteLibraries`, they would hit this issue too.

packages/vite-plugin-svelte/src/utils/options.ts
```diff
@@ -211,6 +211,24 @@ export function buildExtraViteConfig(
 	// @ts-ignore
 	extraViteConfig.ssr = buildSSROptionsForSvelte(svelteDeps, options, config);
 
+	if (configEnv.command === 'serve') {
+		// during dev, we have to externalize transitive cjs only dependencies, see https://github.com/sveltejs/vite-plugin-svelte/issues/281
+		const optimizedTransitiveDeps = extraViteConfig.optimizeDeps?.include
+			?.filter((x) => x.includes('>'))
+			.map((x) => x.substring(x.lastIndexOf('>') + 1).trim());
+		if (optimizedTransitiveDeps?.length) {
+			// @ts-expect-error ssr still flagge in vite
```
benmccann, line 220: missing a d in "flagge"?  it's repeated several times below as well

packages/vite-plugin-svelte/src/utils/options.ts
dominikg: moved it and cleaned up. not sure i follow the prebundleSvelteLibraries problem.

packages/vite-plugin-svelte/src/utils/options.ts
dominikg: yes, fixed. Thanks

dominikg:
is externalizing everything ok? if that was the case, why is vite doing this dance with noExternal in the first place?

benmccann:
I believe it should be okay. At this point, I think the dance really isn't necessary and is just legacy that no one's taken the time to cleanup yet. Pretty much everything is already externalized and I'm not sure it would have any user visible changes. I think it's really just a single test that needs to be removed or updated. The failing test is `optimize-missing-deps`, which I'm really confused about because I'm not sure how anything can happen one way or another to a dependency that's missing. I don't really understand what it's supposed to be testing. At most I think we'd just need to special case missing dependencies and everything else can be externalized

packages/vite-plugin-svelte/src/utils/options.ts
bluwy: If `prebundleSvelteLibraries` is set to `true`, `buildOptimizeDepsForSvelte()` will early return here:

https://github.com/sveltejs/vite-plugin-svelte/blob/681002cf52cea1e0e5b3c8439151dcffe26aa5c0/packages/vite-plugin-svelte/src/utils/options.ts#L244-L252

Skipping the below code that handles adding `parent > child` to `optimizeDeps.include` so there won't be any `>` syntax to handle in configuring `ssr.external` later on, which brings up the initial issue where nested deps (e.g. `child`) weren't added to `ssr.external` when it should.

To showcase this visually with the current PR's changes:

If `prebundleSvelteLibraries: false` (default)
  - `optimizeDeps.include: ["svelte-dep > cjs-dep"]`
  - `optimizeDeps.exclude: ["svelte-dep"]`
  - `ssr.noExternal: ["svelte-dep"]`
  - `ssr.external: ["cjs-dep"]`

If `prebundleSvelteLibraries: true`
  - `optimizeDeps.include: []`
  - `optimizeDeps.exclude: []`
  - `ssr.noExternal: ["svelte-dep"]`
  - `ssr.external: []` <- should have `cjs-dep`
  
So we have to reconstruct the `ssr.external` without relying on `optimizeDeps.include` 😵

bluwy:
I think it would be great if we can find a way to externalize everything by default. I checked the failing test and it seems like Jest is [interfering it](https://github.com/vitejs/vite/runs/5009714321?check_suite_focus=true#step:9:177) with it's own transformation. Perhaps the CJS/ESM check Vite has avoided this specific edge case, but it looks like a Jest issue to me if things are working locally without Jest. I haven't looked too deep yet.

benmccann:
Ah, if Jest was doing a transform that would explain a lot because it looked like I was getting code that was different than what was on disk. Another thing I noticed is that one of the `package.json` files in that test has ESM code, but is missing `"type": "module"`

bluwy (approved):
LGTM

bluwy (approved):

packages/vite-plugin-svelte/src/utils/options.ts
```diff
@@ -318,14 +317,12 @@ function buildSSROptionsForSvelte(
...
-		if (optimizedTransitiveDeps?.length) {
-			// @ts-expect-error ssr still flagged in vite
-			ssr.external = optimizedTransitiveDeps.filter((x) => !config.ssr?.noExternal?.includes(x));
-		}
+		// during dev, we have to externalize transitive dependencies, see https://github.com/sveltejs/vite-plugin-svelte/issues/281
+		const transitiveDepsOfSvelteLibs = [
+			...new Set(svelteDeps.flatMap((dep) => Object.keys(dep.pkg.dependencies || {})))
+		];
+		// @ts-expect-error ssr still flagged in vite
+		ssr.external = transitiveDepsOfSvelteLibs;
```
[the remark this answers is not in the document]
bluwy: I think it should be safe to fix this here so it works in Vite 2.8 at the meantime. Having this handled in Vite 2.9 ootb would be great though so perhaps later on we can remove this.

PR #294, opened 2022-03-08
feat: correctly prebundle svelte libraries

### LINKED ISSUE #222: Cleanup prebundleSvelteLibraries

bluwy:
Requires Vite 2.9, ~~though I think it works with Vite 2.8 too (with caveats)~~, but may be better to still bump the peer dep. Putting this here for visibility and if others like to try.

Closes #222


.changeset/young-chicken-agree.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sveltejs/vite-plugin-svelte': major
```
bluwy, line 2: Got this warning when I choose "major":
```
🦋  WARNING: Releasing a major version for @sveltejs/vite-plugin-svelte will be its first major release.
🦋  If you are unsure if this is correct, contact the package's maintainers before committing this changeset.
```

Don't think I've seen this before, but maybe it's still safe? Otherwise we can choose minor, and the new release according to semver should still be a breaking change.

dominikg: Seen that before

packages/vite-plugin-svelte/src/index.ts
```diff
@@ -70,7 +70,13 @@ export function svelte(inlineOptions?: Partial<Options>): Plugin {
 		},
 
 		async buildStart() {
-			await handleOptimizeDeps(options, viteConfig);
+			if (!options.experimental.prebundleSvelteLibraries) return;
+			const isSvelteMetadataChanged = await saveSvelteMetadata(viteConfig.cacheDir, options);
+			if (isSvelteMetadataChanged) {
+				// Force Vite to optimize again. Although we mutate the config here, it works because
+				// Vite's optimizer runs after `buildStart()`.
+				viteConfig.server.force = true;
```
dominikg, line 78: Is this safe? What happens when the server is restarted afterwards

packages/vite-plugin-svelte/src/utils/optimizer.ts
```diff
@@ -13,24 +12,28 @@ const PREBUNDLE_SENSITIVE_OPTIONS: (keyof ResolvedOptions)[] = [
...
+	const svelteMetadata = generateSvelteMetadata(options);
+	const svelteMetadataPath = path.resolve(cacheDir, '_svelte_metadata.json');
 
-	const viteMetadataPath = findViteMetadataPath(viteConfig.cacheDir);
-	if (!viteMetadataPath) return;
-
-	const svelteMetadataPath = path.resolve(viteMetadataPath, '../_svelte_metadata.json');
-	const currentSvelteMetadata = JSON.stringify(generateSvelteMetadata(options), (_, value) => {
+	const currentSvelteMetadata = JSON.stringify(svelteMetadata, (_, value) => {
+		// Handle preprocessors
```
dominikg, line 23: Are the function strings eval'd at any point?

packages/vite-plugin-svelte/src/utils/optimizer.ts
```diff
@@ -13,24 +12,28 @@ const PREBUNDLE_SENSITIVE_OPTIONS: (keyof ResolvedOptions)[] = [
...
+		existingSvelteMetadata = await fs.readFile(svelteMetadataPath, 'utf8');
+	} catch {
+		// ignore
 	}
 
-	await optimizeDeps(viteConfig, true);
-	fs.writeFileSync(svelteMetadataPath, currentSvelteMetadata);
+	await fs.mkdir(cacheDir, { recursive: true });
+	await fs.writeFile(svelteMetadataPath, currentSvelteMetadata);
+	return currentSvelteMetadata !== existingSvelteMetadata;
```
dominikg, line 36: Do we need deep object compare here or is it deterministic?

packages/vite-plugin-svelte/src/index.ts
bluwy: Hmm yeah. Looks like if the server restart later on, it would always issue force optimization. Maybe reverting this on `buildEnd()` would fix this. I'll test this later on.

packages/vite-plugin-svelte/src/utils/optimizer.ts
bluwy: I don't think so. They're stringified because preprocessors don't have a `.name` to be keyed. The metadata acts like hash currently, so we don't have to do a `JSON.parse` for this metadata.

packages/vite-plugin-svelte/src/utils/optimizer.ts
bluwy: It should be deterministic since they are strings and we're constructing the metadata object (that is stringified) in a consistent way.

packages/vite-plugin-svelte/src/index.ts
bluwy: I just tested this and the `buildEnd()` thing isn't needed. It looks like Vite re-constructs the config on restart so this change won't be carried on after restarts. With that said, I think this trick should still be safe and there isn't any other way to achieve the behaviour, since Vite 2.9 prebundles dep in a non-blocking way now.

dominikg: Feels hacky still. Is it possible to use server.restart or add api for this to vite?

packages/vite-plugin-svelte/src/utils/optimizer.ts
dominikg: Talked about it on discord. This could have false positives when keys/arrays are reordered but worst case it prebundles again

packages/vite-plugin-svelte/src/index.ts
bluwy: `server.restart` is tricky since we don't have the new resolved config before determining to pass `server.restart(true)` or not. And that would handle the case while the dev server is running and restart. If the user changed the config and then start the dev server, we'd have to checked that too externally.

Perhaps an API vite can provide is `server.forceOptimize` but I'll have to look into that. The trick here is actually borrowed from [here](https://github.com/vitejs/vite/blob/a74bd7ba9947be193bccf636b8918bd1f59e89ae/packages/vite/src/node/server/index.ts#L736-L742).

dominikg: hmm, it feels like this warrants/needs new api in vite. plugins contributing to prebundling/optimization needing to hack things doesn't feel like it's a long term solution. with vite3 coming up maybe it would be a good time to do something about it?

while it may work now it'll break the second someone refactors the logic around vite server.force. We might learn about it in vite-ecosystem-ci before it actually burns users but this creates a coupling i'm not comfortable with :/

bluwy: Yeah. I'll take a look at supporting an official API at the meantime. I have some ideas about it which might work, and having a safe API for plugins to use would be good for vue/marko adoption too.

I'm not sure how long that'll take though, once I got something up I'll check with patak to see if we can ship it soon.

Perhaps we can still merge this given it's behind an experimental flag. Probably won't turn it on for SvelteKit yet, but with the edge cases resolved we may be able to get more feedback. I'll also be keeping an eye in Vite if anyone tries to hurt my hack 😄

dominikg: as long as it stays behind an experimental flag i'm fine with it (: +1:

dominikg (approved):

bluwy:
Merging this in so we don't hold it any longer. Summary: There's [one more thing](https://github.com/sveltejs/vite-plugin-svelte/pull/294#discussion_r841076032) to do to make sure we can force reoptimization of the dev server without a hack.

PR #305, opened 2022-04-03
fix: Improve css sourcemap accuracy

### LINKED ISSUE #300: css sourcemaps
### Describe the problem

Prior to vite 2.9 vite did not provide sourcemaps for css in dev. now it does so we should make sure it works with svelte component styles too.

see what was done for vite-plugin-vue and follow/adapt https://twitter.com/sapphi_red/status/1506272042257059842

### Alternatives considered

keep it as is, with potentially invalid maps.

### Importance

would make my life easier

bfanger:
When using Vite's (experimental) [css.devSourcemap](https://vitejs.dev/config/#css-devsourcemap) feature
The filenames where correct, but the line numbers didn't match.

Changing:
`source: filename`
into
`source: filename?.split(/[/\\]/).pop()`

By changing `source` to only contain the filename instead of the full path fixed the line number issue.

Fixes #300

### DIFF
--- a/packages/vite-plugin-svelte/src/utils/preprocess.ts
+++ b/packages/vite-plugin-svelte/src/utils/preprocess.ts
@@ -11,6 +11,7 @@ import { Preprocessor, PreprocessorGroup, Processed, ResolvedOptions } from './o
 import { TransformPluginContext } from 'rollup';
 import { log } from './log';
 import { buildSourceMap } from './sourcemap';
+import path from 'path';
 
 const supportedStyleLangs = ['css', 'less', 'sass', 'scss', 'styl', 'stylus', 'postcss'];
 
@@ -57,7 +58,7 @@ function createViteStylePreprocessor(config: ResolvedConfig): Preprocessor {
 		)) as TransformResult;
 		// patch sourcemap source to point back to original filename
 		if (transformResult.map?.sources?.[0] === moduleId) {
-			transformResult.map.sources[0] = filename;
+			transformResult.map.sources[0] = path.basename(filename);
 		}
 		return {
 			code: transformResult.code,
@@ -94,7 +95,10 @@ function createInjectScopeEverythingRulePreprocessorGroup(): PreprocessorGroup {
 			s.append(' *{}');
 			return {
 				code: s.toString(),
-				map: s.generateDecodedMap({ source: filename, hires: true })
+				map: s.generateDecodedMap({
+					source: filename ? path.basename(filename) : undefined,
+					hires: true
+				})
 			};
 		}
 	};
--- a/.changeset/thin-swans-talk.md
+++ b/.changeset/thin-swans-talk.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/vite-plugin-svelte': patch
+---
+
+Improved CSS Source Maps when using vite's `css: { devSourcemap: true }`
--- a/packages/playground/kit-demo-app/svelte.config.js
+++ b/packages/playground/kit-demo-app/svelte.config.js
@@ -8,6 +8,9 @@ const config = {
 		// Override http methods in the Todo forms
 		methodOverride: {
 			allowed: ['PATCH', 'DELETE']
+		},
+		vite: {
+			css: { devSourcemap: true }
 		}
 	}
 };


dominikg:
This is a great finding, thanks!

The PR needs to be rebased due to conflicts

packages/vite-plugin-svelte/src/utils/preprocess.ts
```diff
@@ -57,7 +57,7 @@ function createViteStylePreprocessor(config: ResolvedConfig): Preprocessor {
 		)) as TransformResult;
 		// patch sourcemap source to point back to original filename
 		if (transformResult.map?.sources?.[0] === moduleId) {
-			transformResult.map.sources[0] = filename;
+			transformResult.map.sources[0] = filename.split(/[/\\]/).pop() as string;
```
dominikg, line 60: i'd prefer `path.basename` here for readability

packages/vite-plugin-svelte/src/utils/preprocess.ts
```diff
@@ -94,7 +94,7 @@ function createInjectScopeEverythingRulePreprocessorGroup(): PreprocessorGroup {
 			s.append(' *{}');
 			return {
 				code: s.toString(),
-				map: s.generateDecodedMap({ source: filename, hires: true })
+				map: s.generateDecodedMap({ source: filename?.split(/[/\\]/).pop(), hires: true })
```
dominikg, line 97: path.basename too?

dominikg:
some questions

1) do we want to handle configuration automatically
  a) set for svelte when enabled in vite config
  b) set for vite when enabled in svelte config 
  c) warn for conflicts?

2) do we need to patch the output sourcemap  either using https://github.com/vitejs/vite/blob/ace6e93239c85667eaea656cfda770e3269b336f/packages/vite/src/node/plugins/css.ts#L869 directly or some similar approaches.

bfanger:
Rebased and changed from split().pop() to using `path.basename`.

1. Automatic Configuration
This change does not impact any configuration for this plugin.
I should mention that to activate the sourcemap feature you do need to 
add to your vite.config:
```
css: { devSourcemap: true }
``` 
 or when using sveltekit to your svelte.config:
```
  kit: { 
    vite: {
      css: { devSourcemap: true }
    },
    ...
```
As you can see this a configuration option for vite, not the plugin.


2. patching sourcemap?
i think that step is already handled by svelte.preprocess & svelte.compile 
There might be some additional work needed for the svelte-preprocess package, but that's another repo.

I've also tested some sass using the `experimental:{ useVitePreprocess: true },` and the generated sourcemap looked fine.

To debug i've used https://evanw.github.io/source-map-visualization/ and pasted the results i logged in packages/vite-plugin-svelte/src/utils/compile.ts using:
```
compiled.css.map && console.log(`\n${filename}\n\n${compiled.css.code}\n/*# sourceMappingURL=${compiled.css.map.toUrl()} */\n\n`) 
```

dominikg:
Sorry my comment above wasn't very clear.

>    I should mention that to activate the sourcemap feature you do need to
>    add to your vite.config:
> 
> ```
> css: { devSourcemap: true }
> ```

vite-plugin-svelte can set `devSourcemap: true` automatically in the config hook so the user doesn't have to do it. This would work for both vite+svelte and sveltekit ootb. 

It could be added here: https://github.com/sveltejs/vite-plugin-svelte/blob/bdebf4e53c47546d41e2abe5f2817bcaa8707f35/packages/vite-plugin-svelte/src/utils/options.ts#L202

unfortunately there are 2 options `svelte.compilerOptions.enableSourcemap.css` and `vite.css.devSourcemap` so if either of them is false in user config, the other should be too.

bfanger:
I would not auto enable it just yet, i've seen this _experimental_ feature throw javascript exceptions during build/dev.
(with an older complex sass setup)

 I'd assume vite will enable the feature by default at some point.

packages/vite-plugin-svelte/src/utils/preprocess.ts
```diff
@@ -94,7 +95,9 @@ function createInjectScopeEverythingRulePreprocessorGroup(): PreprocessorGroup {
 			s.append(' *{}');
 			return {
 				code: s.toString(),
-				map: s.generateDecodedMap({ source: filename, hires: true })
+				map: filename
```
dominikg, line 98: i think filename is always set here, but just to be sure it would be better to always generate the map but leave source undefined ` s.generateDecodedMap({ source: filename ? path.basename(filename) : undefined , hires: true })`

dominikg:
> I would not auto enable it just yet, i've seen this _experimental_ feature throw javascript exceptions during build/dev. (with an older complex sass setup)
> 
> I'd assume vite will enable the feature by default at some point.

we can handle the config later in a separate PR, true. although i do lean on early adpoting things that improve the dev experience. and missing css sourcemaps has been a sore spot.

dominikg:
final step missing before merge would be a changeset `pnpm changeset` create a patch for vite-plugin-svelte with a small note

dominikg (approved):

bfanger:
👍 done!

bluwy (approved):

dominikg:
Released. Thanks!

https://github.com/sveltejs/vite-plugin-svelte/releases/tag/%40sveltejs/vite-plugin-svelte%401.0.0-next.41

PR #324, opened 2022-05-06
fix: import ts with .js extension

### LINKED ISSUE #323: Could not resolve .js import from ts svelte component
### Describe the bug

When `type: module` is used in package.json, the specification says that any import should have an extension. Following convention, if we add .js import for a typescript file, it is not able to resolve.

### Reproduction

```svelte
<script lang="ts">
import foo from './some-typescript-file.js';
console.log(foo);
</script>
```

This works with a regular .ts file, but not a svelte file using ts context.

### Logs

```shell
Errors:

Could not resolve './some-typescript-file.js' from App.svelte
```
```


### System Info

```shell
System:
    OS: macOS 10.15.7
    CPU: (16) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 48.13 GB / 64.00 GB
    Shell: 5.7.1 - /bin/zsh
  Binaries:
    Node: 16.15.0 - ~/.nvm/versions/node/v16.15.0/bin/node
    Yarn: 1.22.18 - /usr/local/bin/yarn
    npm: 8.5.5 - ~/.nvm/versions/node/v16.15.0/bin/npm
  Browsers:
    Chrome: 100.0.4896.127
    Firefox: 98.0.2
    Safari: 15.4
  npmPackages:
    @sveltejs/vite-plugin-svelte: 1.0.0-next.42 => 1.0.0-next.42 
    svelte: 3.48.0 => 3.48.0 
    vite-system:  2.9.6
```

### Severity

blocking all usage of vite-plugin-svelte

bluwy:
Fix #323 

Requires Vite 2.9.8 to work, but not strictly needed as peer dep.


packages/vite-plugin-svelte/src/utils/preprocess.ts
```diff
@@ -194,6 +194,16 @@ export function addExtraPreprocessors(options: ResolvedOptions, config: Resolved
 	}
 }
 
+export function createSpyLangPreprocessor(map: Map<string, string>): PreprocessorGroup {
+	return {
+		script({ attributes, filename }) {
+			if (filename && attributes.lang) {
+				map.set(filename, attributes.lang as string);
```
dominikg, line 201: what happens if lang is pug? or there are a mixture of different lang attributes on module and regular script?

bluwy: Removed the preprocessor to a simple regex. If there's different lang, it would pick the first defined one as the source of truth, which should be good enough for ts support.

packages/vite-plugin-svelte/src/utils/compile.ts
```diff
@@ -6,6 +6,8 @@ import { SvelteRequest } from './id';
 import { safeBase64Hash } from './hash';
 import { log } from './log';
 
+const scriptLangRE = /<script.+lang="(.+)".*>/;
```
dominikg, line 9:
```suggestion
const scriptLangRE = /<script [^>]*lang=[\"']?([^\"' >]+)[\"']?[^>]*>/;
```

avoids greed and allows all quote styles:
```
<script lang="ts" foo="bar">
<script lang='ts'>
<script lang=ts>
<script lang='ts' foo='bar'>
<script lang=ts foo=bar>
```

one weakness is it would not match a quoted lang value that contains a space `lang="why do that"`

bluwy: Always great to be more robust 👍

dominikg (approved):

PR #338, opened 2022-05-15
fix(inspector): handle autocomplete keydown event

MathiasWP:
Chrome triggers a "keydown" event when autocompleting in forms, which triggers a `event.getModifierState is not a function` error.
This change makes sure that only keydown events with a defined key are checked.

Related issue from another GitHub repo: https://github.com/paperjs/paper.js/issues/1398

Screenshot of bug from my project:
Screenshot 2022-05-15 at 13 38 51


bluwy (approved):

bluwy:
Thanks!

Issue #345, opened 2022-05-24
Ignores compilerOptions.generate


ghostdevv:
### Describe the bug

@jakobrosenberg and I are trying to generate two builds of a vite app and we can't get v-p-s to always force ssr to false

Jake came up with the following workaround:
https://github.com/roxiness/routify.dev/commit/e8638a8ca48fd9cf503fdd60a608217853b27720#diff-58e6f63d87181b1c6a8cb6e5f1691df04aa32854456efcd52ca71c8541375d26R14

### Reproduction

We have these scripts in the package.json:

```json
{
    "build:client": "vite build --outDir dist/client",
    "build:server": "vite build --outDir dist/server --ssr src/main.js",
}
```

and trying to force ssr to always be false for svelte here:

```js
    svelte({
      compilerOptions: { generate: "dom", format: "esm" },
```

### Logs

```shell
https://gist.github.com/ghostdevv/0850dada3de50545b3b2e556f688afe4
```

### System Info

```shell
System:
    OS: Linux 5.17 undefined
    CPU: (12) x64 AMD Ryzen 5 3600 6-Core Processor
    Memory: 430.21 MB / 15.62 GB
    Container: Yes
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 16.9.1 - ~/.nvm/versions/node/v16.9.1/bin/node
    Yarn: 1.22.18 - ~/.nvm/versions/node/v16.9.1/bin/yarn
    npm: 7.24.2 - ~/.nvm/versions/node/v16.9.1/bin/npm
  Browsers:
    Brave Browser: 101.1.38.111
    Chromium: 101.0.4951.54
    Firefox: 100.0
  npmPackages:
    @sveltejs/vite-plugin-svelte: ^1.0.0-next.44 => 1.0.0-next.44 
    svelte: ^3.48.0 => 3.48.0 
    vite: ^2.9.9 => 2.9.9
```

### Severity

annoyance

dominikg:
generate is indeed hardcoded: https://github.com/sveltejs/vite-plugin-svelte/blob/main/packages/vite-plugin-svelte/src/utils/compile.ts#L24

If you don't want an ssr build and instead render the result yourself with jsdom, not passing `--ssr` to vite build could work: 
`"build:server": "vite build --outDir dist/server src/main.js"`

bluwy:
[...]

PR #346, opened 2022-05-26
fix: Remove user-set values for essential compilerOptions and log a warning

dominikg:
see https://github.com/sveltejs/vite-plugin-svelte/issues/345#issuecomment-1138428492

this does not apply to dynamicCompileOptions so that remains  an escape-hatch for the brave.

### DIFF
--- a/.changeset/three-icons-behave.md
+++ b/.changeset/three-icons-behave.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/vite-plugin-svelte': patch
+---
+
+Remove user-set values for essential compilerOptions generate, format, cssHash and filename and log a warning
--- a/.changeset/three-icons-behave.md
+++ b/.changeset/three-icons-behave.md
@@ -2,4 +2,4 @@
 '@sveltejs/vite-plugin-svelte': patch
 ---
 
-Remove user-set values for essential compilerOptions generate, format, cssHash and filename and log a warning
+Remove user-specified values for essential compilerOptions generate, format, cssHash and filename and log a warning
--- a/packages/vite-plugin-svelte/src/utils/options.ts
+++ b/packages/vite-plugin-svelte/src/utils/options.ts
@@ -176,7 +176,10 @@ function enforceOptionsForProduction(options: ResolvedOptions) {
 }
 
 function removeIgnoredOptions(options: ResolvedOptions) {
-	const ignoredCompilerOptions = ['generate', 'cssHash', 'format', 'filename'];
+	const ignoredCompilerOptions = ['generate', 'format', 'filename'];
+	if (options.hot && options.emitCss) {
+		ignoredCompilerOptions.push('cssHash');
+	}
 	const passedCompilerOptions = Object.keys(options.compilerOptions || {});
 	const passedIgnored = passedCompilerOptions.filter((o) => ignoredCompilerOptions.includes(o));
 	if (passedIgnored.length) {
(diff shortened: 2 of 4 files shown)


packages/vite-plugin-svelte/src/utils/options.ts
```diff
@@ -177,6 +175,23 @@ function enforceOptionsForProduction(options: ResolvedOptions) {
 	}
 }
 
+function removeIgnoredOptions(options: ResolvedOptions) {
+	const ignoredCompilerOptions = ['generate', 'cssHash', 'format', 'filename'];
+	const passedCompilerOptions = Object.keys(options.compilerOptions || {});
+	const passedIgnored = passedCompilerOptions.filter((o) => ignoredCompilerOptions.includes(o));
+	if (passedIgnored.length) {
+		log.warn(
+			`The following svelte compilerOptions are controlled by vite-plugin-svelte and essential to it's functionality. User-Set values are ignored. Please remove them from your configuration: ${passedIgnored.join(
```
Conduitry, line 184:
```suggestion
			`The following Svelte compilerOptions are controlled by vite-plugin-svelte and essential to its functionality. User-specified values are ignored. Please remove them from your configuration: ${passedIgnored.join(
```

dominikg: thanks

Conduitry (approved):

packages/vite-plugin-svelte/src/utils/options.ts
```diff
@@ -177,6 +175,23 @@ function enforceOptionsForProduction(options: ResolvedOptions) {
...
 }
 
+function removeIgnoredOptions(options: ResolvedOptions) {
+	const ignoredCompilerOptions = ['generate', 'cssHash', 'format', 'filename'];
+	const passedCompilerOptions = Object.keys(options.compilerOptions || {});
+	const passedIgnored = passedCompilerOptions.filter((o) => ignoredCompilerOptions.includes(o));
+	if (passedIgnored.length) {
+		log.warn(
+			`The following Svelte compilerOptions are controlled by vite-plugin-svelte and essential to its functionality. User-specified values are ignored. Please remove them from your configuration: ${passedIgnored.join(
+				', '
```
bluwy, line 185: Was initially thinking to remove the props from the `CompilerOptions` types directly, but maybe we can do that and the errors too. (Though I feel like we should lean on types as we only need one line of defence)

.changeset/three-icons-behave.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sveltejs/vite-plugin-svelte': patch
+---
+
+Remove user-set values for essential compilerOptions generate, format, cssHash and filename and log a warning
```
bluwy, line 5:
```suggestion
Remove user-specified values for essential compilerOptions generate, format, cssHash and filename and log a warning
```

Maybe update here too for consistency

packages/vite-plugin-svelte/src/utils/options.ts
dominikg: We should rather get those from svelte. Maybe we can use omit?

packages/vite-plugin-svelte/src/utils/options.ts
```diff
@@ -177,6 +175,23 @@ function enforceOptionsForProduction(options: ResolvedOptions) {
 	}
 }
 
+function removeIgnoredOptions(options: ResolvedOptions) {
+	const ignoredCompilerOptions = ['generate', 'cssHash', 'format', 'filename'];
```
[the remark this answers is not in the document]
dominikg: Good call

Issue #348, opened 2022-05-29
Use `mtime` as cache busting parameter


benmccann:
### Describe the problem

The config file is currently loaded with `Date.now()` as a cache-busting parameter. If `mtime` were instead used then SvelteKit could load the config with the same cache-busting parameter without us having to pass the config around which would simplify some APIs (see https://github.com/sveltejs/kit/pull/5094).

### Describe the proposed solution

Use `mtime`

### Alternatives considered

1. the user imports the config
2. duplicate config loading
3. split the config file between svelte.config.js and vite.config.js
4. plugin intercommunication (though I don't think this would actually work)
5. await plugin creation function

### Importance

would make my life easier

PR #356, opened 2022-05-31
Use mtime as timestamp when loading config file

benmccann:
Closes https://github.com/sveltejs/vite-plugin-svelte/issues/348

### DIFF
--- a/packages/vite-plugin-svelte/src/utils/load-svelte-config.ts
+++ b/packages/vite-plugin-svelte/src/utils/load-svelte-config.ts
@@ -22,7 +22,8 @@ export const knownSvelteConfigNames = [
 // also use timestamp query to avoid caching on reload
 const dynamicImportDefault = new Function(
 	'path',
-	'return import(path + "?t=" + Date.now()).then(m => m.default)'
+	'timestamp',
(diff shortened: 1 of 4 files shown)


packages/vite-plugin-svelte/src/utils/load-svelte-config.ts
```diff
@@ -38,7 +39,10 @@ export async function loadSvelteConfig(
 		// try to use dynamic import for svelte.config.js first
 		if (configFile.endsWith('.js') || configFile.endsWith('.mjs')) {
 			try {
-				const result = await dynamicImportDefault(pathToFileURL(configFile).href);
+				const result = await dynamicImportDefault(
+					pathToFileURL(configFile).href,
+					fs.statSync(configFile).mtime
```
Conduitry, line 44:
```suggestion
					fs.statSync(configFile).mtimeMs
```

maybe? Just using `mtime` will get a `Date` instance I believe, which will then be formatted via whatever localization you happen to get with `.toString()`, which is going to contain spaces and other characters that should be URL encoded.

`mtimeMs`, on the other hand, is a decimal number. Might not matter, but it seems a bit tidier.

dominikg (approved):

Issue #362, opened 2022-06-03
Build error: Received '\x00__vite-browser-external?commonjs-proxy'


niczero:
### Describe the bug

I am seeing a build error when working on a project of Inertia + Svelte using Vite.  The code works fine in development mode with HMR, but shows this error when doing a production build.
```
2022-06-03T07:28:49.338Z vite:vite-plugin-svelte error trying to resolve __vite-browser-external from __vite-browser-external?commonjs-proxy via package.json svelte field  TypeError [ERR_INVALID_ARG_VALUE]: The argument 'filename' must be a file URL object, file URL string, or absolute path string. Received '\x00__vite-browser-external?commonjs-proxy'
    at new NodeError (node:internal/errors:372:5)
    at createRequire (node:internal/modules/cjs/loader:1220:13)
    at resolveViaPackageJsonSvelte ($DIR/vite-plugin-svelte-bug/node_modules/@sveltejs/vite-plugin-svelte/dist/index.cjs:1476:59)
    at Object.resolveId ($DIR/vite-plugin-svelte-bug/node_modules/@sveltejs/vite-plugin-svelte/dist/index.cjs:1728:28)
    at $DIR/vite-plugin-svelte-bug/node_modules/rollup/dist/shared/rollup.js:22974:37 {
  code: 'ERR_INVALID_ARG_VALUE'
}
```
Is the error benign?
Is the root cause on the Inertia side?

### Reproduction

I have supplied a minimal repro repo at https://github.com/niczero/vite-plugin-svelte-bug.

The package.json is:
```
{
  "dependencies": {
    "@inertiajs/inertia": "^0.11.0",
    "@inertiajs/inertia-svelte": "^0.8.0",
    "svelte": "^3.48.0"
  },
  "devDependencies": {
    "@sveltejs/vite-plugin-svelte": "^1.0.0-next.47",
    "svelte-loader": "^3.1.3",
    "svelte-preprocess": "^4.10.6",
    "vite": "^2.9.9"
  },
  "name": "vite-build-bug",
  "private": true,
  "scripts": {
    "build": "vite build --debug"
  }
}
```

### Logs

```shell
Log from `vite build --debug`:

> build
> vite build --debug

2022-06-03T07:28:48.601Z vite:config bundled config file loaded in 74.53ms
[...]

PR #363, opened 2022-06-03
fix: ignore __vite-browser-external when trying to resolve svelte packages

dominikg:
this removes a confusing error log but doesn't fix the underlying issue which is in vite itself. see #362

### DIFF
--- a/.changeset/stupid-melons-try.md
+++ b/.changeset/stupid-melons-try.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/vite-plugin-svelte': patch
+---
+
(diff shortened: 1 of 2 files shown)


bluwy:
Hmm I thought we had wrapped a try-catch to prevent edge-cases like this from breaking dev/build 🤔 And with #362 already closed, do we still need this?

dominikg:
Yes, to prevent the confusing error log

bluwy:
Ah I see it's just the debug log from the issue, build was fine. Wouldn't hurt to add another check then 👍

bluwy (approved):

bluwy (approved):

PR #368, opened 2022-06-07
feat: automatic handling for kit.browser.hydrate config

dominikg:
This isn't really testable until kit stops passing `configFile: false` in inline options.

Manually ignoring that flag and removing the inline kit options show it works though.


packages/vite-plugin-svelte/src/utils/options.ts
```diff
@@ -195,6 +197,22 @@ function removeIgnoredOptions(options: ResolvedOptions) {
 	}
 }
 
+function setKitOptions(options: ResolvedOptions) {
```
benmccann, line 200: it's not actually the Kit options that are being set here, but the compiler options. maybe it should be called `setUsingKitOptions` or something

benmccann (approved):
thanks!!

packages/vite-plugin-svelte/src/utils/options.ts
dominikg: thanks, upated and added a small comment

benmccann (approved):

PR #389, opened 2022-07-12
refactor: move plugin options to "vitePlugin" in svelte.config.js

dominikg:
To avoid problems with future name clashes we introduce `vitePlugin` namespace in svelte.config.js.

svelte options that are of global interest, eg `compilerOptions`, `preprocess` and `extensions` remain in the config root.

All other options mentioned in https://github.com/sveltejs/vite-plugin-svelte/blob/main/docs/config.md#plugin-options are moved.

Additional validation/errors have been introduced to avoid invalid options and inform the user.

### DIFF
--- a/.changeset/healthy-worms-double.md
+++ b/.changeset/healthy-worms-double.md
@@ -9,10 +9,10 @@ update your svelte.config.js and wrap [plugin options](https://github.com/svelte
 ```diff
 // svelte.config.js
 
-  compilerOptions:{...},
-  preprocess:{...},
-  extensions:[...],
-  kit:{},
+  compilerOptions: {...},
+  preprocess: {...},
+  extensions: [...],
+  kit: {},
 + vitePlugin: {
    // include, exclude, emitCss, onwarn, hot, ignorePluginPreprocessors, disableDependencyReinclusion, experimental
 + }
--- a/.changeset/healthy-worms-double.md
+++ b/.changeset/healthy-worms-double.md
@@ -12,8 +12,9 @@ update your svelte.config.js and wrap [plugin options](https://github.com/svelte
   compilerOptions: {...},
   preprocess: {...},
   extensions: [...],
+  onwarn: () => {...},
   kit: {},
 + vitePlugin: {
-   // include, exclude, emitCss, onwarn, hot, ignorePluginPreprocessors, disableDependencyReinclusion, experimental
+   // include, exclude, emitCss, hot, ignorePluginPreprocessors, disableDependencyReinclusion, experimental
 + }
 ```
--- a/.changeset/healthy-worms-double.md
+++ b/.changeset/healthy-worms-double.md
@@ -1,0 +1,19 @@
+---
+'@sveltejs/vite-plugin-svelte': major
+---
+
+move plugin options in svelte.config.js into "vitePlugin"
+
+update your svelte.config.js and wrap [plugin options](https://github.com/sveltejs/vite-plugin-svelte/blob/main/docs/config.md#plugin-options) with `vitePlugin`
+
+```diff
+// svelte.config.js
+
+  compilerOptions:{...},
+  preprocess:{...},
+  extensions:[...],
+  kit:{},
++ vitePlugin: {
+   // include, exclude, emitCss, onwarn, hot, ignorePluginPreprocessors, disableDependencyReinclusion, experimental
++ }
+```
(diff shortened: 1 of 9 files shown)


ghostdevv:
Are you planning on supporting both for a while? there have been so many breaking changes in svelte projects recently it might be nice to have a break :sweat_smile:

dominikg:
No, this is breaking. But it should be the last one before vite-plugin-svelte 1.0.0

docs/config.md
```diff
@@ -197,7 +204,10 @@ A [picomatch pattern](https://github.com/micromatch/picomatch), or array of patt
 
 These options are considered experimental and breaking changes to them can occur in any release! Specify them under the `experimental` option.
 
+Either in Vite config
```
bluwy, line 207:
```suggestion
Either in Vite config:
```

docs/config.md
```diff
@@ -209,6 +219,19 @@ export default defineConfig({
 });
 ```
 
+or in svelte config
```
bluwy, line 222:
```suggestion
or in Svelte config:
```

packages/e2e-tests/configfile-custom/svelte.config.custom.cjs
```diff
@@ -1,4 +1,6 @@
 console.log('custom svelte config loaded cjs')
 module.exports = {
-	emitCss: false
+	vitePlugin:{
```
bluwy, line 3:
```suggestion
	vitePlugin: {
```

packages/e2e-tests/configfile-custom/svelte.config.custom.mjs
```diff
@@ -1,4 +1,6 @@
 console.log('custom svelte config loaded mjs')
 export default {
-	emitCss: false
+	vitePlugin:{
```
bluwy, line 3:
```suggestion
	vitePlugin: {
```

.changeset/healthy-worms-double.md
```diff
@@ -0,0 +1,19 @@
...
+
+update your svelte.config.js and wrap [plugin options](https://github.com/sveltejs/vite-plugin-svelte/blob/main/docs/config.md#plugin-options) with `vitePlugin`
+
+```diff
+// svelte.config.js
+
+  compilerOptions:{...},
+  preprocess:{...},
+  extensions:[...],
+  kit:{},
```
bluwy, line 15:
```suggestion
  compilerOptions: {...},
  preprocess: {...},
  extensions: [...],
  kit: {},
```

bluwy (commented):
Some spacing nits, but the src looks good to me. Maybe we also want to export a new type for `vitePlugin`? Currently we have `Options` that's for passing the the plugin directly.

bluwy (approved):

ganigeorgiev:
@bluwy, @dominikg Is this change available in the recent v1.0.1 release?

When I wrap my options under `vitePlugin` as per the release note, like this:
```
// vite.config.js
export default defineConfig({
    plugins: [
        svelte({
            vitePlugin: {
                experimental: {
                    useVitePreprocess: true,
                },
            },
        }),
    ],
    ...
});
```

I'm seeing the following warning in the console:
```
[vite-plugin-svelte] invalid plugin options "vitePlugin" in inline config
{ vitePlugin: { experimental: { useVitePreprocess: true } } }
```

dominikg:
inline in vite config you do not wrap them. Only in svelte.config.js

ganigeorgiev:
Ah, sorry. Yes, I've just read https://github.com/sveltejs/vite-plugin-svelte/blob/main/docs/config.md#experimental-options mentioned that this change is only for svelte.config.js.

This is kind of confusing to be honest and would be better to be consistent. Furthermore the example at the top of the page has this:
```js
// vite.config.js
export default defineConfig({
  plugins: [
    svelte({
      configFile: false
      // your svelte config here
    })
  ]
});
```
where I thought because of `// your svelte config here` it should support the same props.

dominikg:
Imho vitePlugin would be redundant inside vite.config.js inline options. It is used in svelte.config.js because that file isn't exclusive for vite-plugin-svelte. 
The inline options in vite.config.js on the other hand are ours to define and use.

ganigeorgiev:
@dominikg Fair enough, I agree. It is just a little confusing and some people may stumble on this, but there is also validation for the config, so I think we are OK.

PR #411, opened 2022-08-17
tests: update to latest svelte-kit and vite

dominikg:


packages/playground/kit-demo-app/package.json
```diff
@@ -1,27 +1,26 @@
...
-    "typescript": "^4.7.4",
-    "vite": "^3.0.7"
-  },
-  "type": "module",
-  "dependencies": {
-    "@fontsource/fira-mono": "^4.5.9",
-    "cookie": "^0.5.0"
-  }
+	"name": "playground-kit-demo-app",
+	"version": "0.0.0",
```
bluwy, line 3: I think we're missing a `"private": true` here

bluwy (commented):
One small issue, but otherwise 🔥 🔥

bluwy (approved):

PR #436, opened 2022-09-14
feat: add drilling and tag display to inspector

### LINKED ISSUE #435: inspector - expand, narrow selection via +/- shortcut on keyboard
### Describe the problem

The inspector is great. The only shortcoming is that sometimes when I select an element, it opens the target component (e.g. Button.svelte) and not the container of that component (e.g. Nav.svelte).
Sometimes it's impossible to select precisely, because there's not even a margin to click in to select (in this case) Nav.svelte.

This feature is literally perfect, and I use it all the time.
This is the only thing missing.

I press my toggleKeyCombo, and while holding it, I strangle the + key on the numeric keypad.
Then the selection should expand from button to parentElement, which is nav.
In the other direction, the - key on the numeric keypad.

### Alternatives considered

It doesn't.

### Importance

would make my life easier

dominikg:
fixes #435 

needs documentation, but where?

svelte-inspector is going to be extracted into a separate package in this monorepo to enable code-sharing with the editor plugin that was demoed on svelte-summit.

@lukaszpolowczyk  i hope you are fine with ArrowUp/Down, that seems more fitting than +/-.

### DIFF
--- a/.changeset/fair-dodos-dance.md
+++ b/.changeset/fair-dodos-dance.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/vite-plugin-svelte': minor
+---
+
+svelte-inspector: use arrow up/down to allow selecting parent/child elements that cannot be moused over (see #435)
--- a/.changeset/light-readers-leave.md
+++ b/.changeset/light-readers-leave.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/vite-plugin-svelte': minor
+---
+
+svelte-inspector: display active element tag
--- a/docs/config.md
+++ b/docs/config.md
@@ -306,6 +306,14 @@ export default {
     toggleKeyCombo?: string;
 
     /**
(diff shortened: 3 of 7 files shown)


bluwy:
I actually suggested arrow keys because not all keyboards has a numeric keypad, I think it'll be nice to have a key that can work universally.

lukaszpolowczyk:
@bluwy Then add support for both arrow keys and the numeric keypad.
I think it will not interfere with anything.

If you do only the arrow keys, I'll shoot myself.

bluwy (approved):
works well!

lukaszpolowczyk:
@bluwy You approve WITHOUT +/-?!

bluwy (approved):

lukaszpolowczyk:
It's a good thing you added the configuration of this shortcut, because I was close to having a heart attack. :O

PR #438, opened 2022-09-15
feat: add keyboard navigation to svelte inspector and improve a11y

dominikg:
this PR follows up on #436 to complete keyboard only navigation and a11y improvements

cc @geoffrich

### DIFF
--- a/packages/vite-plugin-svelte/src/ui/inspector/Inspector.svelte
+++ b/packages/vite-plugin-svelte/src/ui/inspector/Inspector.svelte
@@ -115,7 +115,7 @@
 		active_el = el;
 	}
 
-	function click(event) {
+	function open_editor(event) {
 		if (file_loc) {
 			stop(event);
 			fetch(`/__open-in-editor?file=${encodeURIComponent(file_loc)}`);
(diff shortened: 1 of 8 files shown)


packages/vite-plugin-svelte/src/ui/inspector/Inspector.svelte
```diff
@@ -222,9 +257,11 @@
 			.join('')}`}
 		on:click={() => toggle()}
 		bind:this={toggle_el}
+		aria-label={`${enabled ? 'disable' : 'enable'} svelte-inspector`}
```
geoffrich, line 260: `aria-label` won't have any effect on a div. Can this be a `button` instead?

geoffrich (commented):
One small comment. The announcer stuff _seems_ like it should work, but you'd probably want to test with VoiceOver or something to be sure.

packages/vite-plugin-svelte/src/ui/inspector/Inspector.svelte
dominikg: yes, button is very possible. thanks for noticing

dominikg: is a button now, thank you `all: unset`

dominikg:
i also cleaned up the changeset files, hoping it creates a cleaner changelog for 1.1.0. given inspector is still experimental and is going to move into a separate package at a later stage, should we use patch instead?

bluwy:
I think it'd be fine if we stick as a patch 👍

bluwy (approved):

PR #469, opened 2022-10-17
refactor: remove `@rollup/pluginutils` dependency

bluwy:
Import `createFilter` from `vite` instead (available since Vite 3).

This should autoclose #468


dominikg (approved):

dominikg:
the filter vite `createFilter` creates behaves identical to rollups i believe

bluwy:
Yeah it's a re-export from Rollup 😬

PR #476, opened 2022-10-20
feat: bring `prebundleSvelteLibraries` out of experimental

benmccann:
Hurray!

We can address https://github.com/sveltejs/vite-plugin-svelte/issues/421 in a follow-up since it only makes prebundling less effective, but doesn't break any projects

### DIFF
--- a/packages/vite-plugin-svelte/src/index.ts
+++ b/packages/vite-plugin-svelte/src/index.ts
@@ -85,7 +85,7 @@ export function svelte(inlineOptions?: Partial<Options>): Plugin[] {
 			},
 
 			async buildStart() {
-				if (!options.experimental?.prebundleSvelteLibraries) return;
+				if (!options.prebundleSvelteLibraries) return;
 				const isSvelteMetadataChanged = await saveSvelteMetadata(viteConfig.cacheDir, options);
 				if (isSvelteMetadataChanged) {
 					// Force Vite to optimize again. Although we mutate the config here, it works because
--- a/packages/vite-plugin-svelte/src/index.ts
+++ b/packages/vite-plugin-svelte/src/index.ts
@@ -67,6 +67,13 @@ export function svelte(inlineOptions?: Partial<Options>): Plugin[] {
 				}
 				// @ts-expect-error temporarily lend the options variable until fixed in configResolved
 				options = await preResolveOptions(inlineOptions, config, configEnv);
+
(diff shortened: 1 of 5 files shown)


packages/vite-plugin-svelte/src/index.ts
```diff
@@ -85,7 +85,7 @@ export function svelte(inlineOptions?: Partial<Options>): Plugin[] {
 			},
 
 			async buildStart() {
-				if (!options.experimental?.prebundleSvelteLibraries) return;
+				if (!options.prebundleSvelteLibraries) return;
```
bluwy, line 88: When resolving the v-p-s options, should we auto-set `options.prebundleSvelteLibraries` if `options.experimental.prebundleSvelteLibraries` is set, and a warning? Not necessarily needed since it was an experimental option, but would be a nice-to-have for existing users who turn this on.

benmccann: yes, good idea. I've added that to the `config` hook

bluwy (approved):
Added a commit to refactor a bit of stuff 😅 but this looks good! Waiting for dominik's approval too just in case.

packages/vite-plugin-svelte/src/utils/options.ts
```diff
@@ -288,6 +290,15 @@ function addSvelteKitOptions(options: ResolvedOptions) {
 	}
 }
 
+function handleDeprecatedOptions(options: ResolvedOptions) {
+	if ((options.experimental as any)?.prebundleSvelteLibraries) {
+		options.prebundleSvelteLibraries = (options.experimental as any)?.prebundleSvelteLibraries;
+		log.warn(
```
dominikg, line 296: This should be thrown as an error. Otherwise people update v-p-s and suddenly svelte dependencies are no longer prebundled and if they don't check warnings that might have scrolled by quickly, they'll miss it.

dominikg:
If you feel comfortable promoting it out of experimental i'm ok with it too. One thing that can be a bit of a pain still is that if it discovers new svelte components while navigating an app during dev, prebundling runs again to include them unless you add all of them to optimizeDeps.include (which has to be done explicitly because there is no glob support in that (why not?)

bluwy:
I think it would only re-prebundle when it discovers a new deep import, but this PR/feature should bring it closer to a Vite issue so maybe we can get glob support for Vite 4 too 🤔 Ref: https://github.com/vitejs/vite/issues/5419

PR #478, opened 2022-10-26
refactor: use vitefu to handle svelte libraries

bluwy:
I may have butchered a few debug logs but I'm not sure if there were helpful recently 😬 Tried to preserve most of the code structure as before so it's easier to review.


dominikg:
Is vitefu using the exact same createRequire approach to resolve package.json paths? if not this could be a breaking change if it ends up resolving something differently.

packages/vite-plugin-svelte/src/utils/options.ts
```diff
@@ -387,82 +422,17 @@ function buildOptimizeDepsForSvelte(
...
-		...Array.from(new Set(svelteDeps.map((s) => s.name))).filter(
-			(x) => !config.ssr?.external?.includes(x)
-		)
-	);
-	const ssr = {
-		noExternal,
-		external: [] as string[]
-	};
-
-	if (options.isServe) {
```
dominikg, line 452: removing this is a different change than adding vitefu?

bluwy: `vitefu` also adds the relevant packages depending on serve or build, which also covers this part.

bluwy:
Currently the only difference is that vitefu uses `import-meta-resolve` rather than `createRequire` to resolve ESM-only packages. Problem: doesn't work with yarn pnp - https://github.com/svitejs/vitefu/issues/2

Other than that, I don't think there's anything else different. The naming conventions such as `component-library` and `js-library` are now `framework-pkg` and `semi-framework-pkg` to keep it agnostic.

packages/vite-plugin-svelte/src/utils/options.ts
```diff
@@ -342,10 +360,30 @@ export function buildExtraViteConfig(
...
 			}
 		};
+	} else {
+		if (options.disableDependencyReinclusion !== true) {
+			const disabledReinclusions = options.disableDependencyReinclusion || [];
+			if (disabledReinclusions.length > 0) {
+				log.debug(`not reincluding transitive dependencies of`, disabledReinclusions);
+			}
+			const transitiveDepsToInclude = depsConfig.optimizeDeps.include.filter((dep) =>
+				disabledReinclusions.every((disabled) => !dep.includes(disabled))
```
dominikg, line 370: What's tested here exactly? The every not could be a some instead and why string.includes instead of equal s?

packages/vite-plugin-svelte/src/utils/options.ts
```diff
@@ -342,10 +360,30 @@ export function buildExtraViteConfig(
...
+		extraViteConfig.optimizeDeps.include?.push(...depsConfig.optimizeDeps.include);
+		extraViteConfig.optimizeDeps.exclude?.push(...depsConfig.optimizeDeps.exclude);
 	}
 
-	// @ts-ignore
-	extraViteConfig.ssr = buildSSROptionsForSvelte(svelteDeps, options, config, extraViteConfig);
+	// ssr config
+	extraViteConfig.ssr = buildSSROptionsForSvelte(config);
+	// @ts-expect-error `noExternal` is 100% an array
+	extraViteConfig.ssr?.noExternal?.push(...depsConfig.ssr.noExternal);
```
dominikg, line 385: the push only happens when extraViteConfin.ssr exists, otherwise depsConfig values will be ignored. Is that correct?

packages/vite-plugin-svelte/src/utils/options.ts
dominikg: Not sure vitefu should make those decision s for vps. My understanding was vitefu provides primitives, otherwise difference s in framework needs can cause problems. Eg a fix for svelte could break solid and vice verca

bluwy: These decisions are required for all frameworks currently though, that's the goal of vitefu, to handle all edge cases within Vite that would work by default. We can postprocess the output if we want to take over with some things.

I don't think a fix for Svelte would break solid (quite depend on what we're fixing), but if that fix would end up benefiting solid/astro along the way, that would be great too.

packages/vite-plugin-svelte/src/utils/options.ts
bluwy: If the user disable reinclusions for some packages, we filter those out from `depsConfig.optimizeDeps.include`. It's using `include` since the specifiers could have `>` for deep optimize.

Do you mean it could be this? 
```suggestion
				!disabledReinclusions.some((disabled) => dep.includes(disabled))
```

packages/vite-plugin-svelte/src/utils/options.ts
bluwy: `buildSSROptionsForSvelte` would ensure `ssr.noExternal` always exist, but I can probably refactor this to make the intent clearer.

packages/vite-plugin-svelte/src/utils/options.ts
dominikg: yes for every vs some, but the `dep.includes` part is dangerous as it could hit false positives when one dependency name is completely contained in another one, eg `svelte-foo` and `not-svelte-foo-but-close`

packages/vite-plugin-svelte/src/utils/options.ts
dominikg: if it is always set, the `?` operator should not be there.

packages/vite-plugin-svelte/src/utils/options.ts
bluwy: ah yeah i didn't thought about that. good call, we should probably fix that

packages/vite-plugin-svelte/src/utils/options.ts
```diff
@@ -323,12 +322,56 @@ export function buildExtraViteConfig(
...
+	} else if (Array.isArray(options.disableDependencyReinclusion)) {
+		const disabledDeps = options.disableDependencyReinclusion;
+		const isDisabled = (dep: string) => {
+			// `crawlFrameworkPkgs` always return `{dep} > {something}` format for deep includes
+			return disabledDeps.some((disabledDep) => dep.includes(`${disabledDep} >`));
+		};
+		depsConfig.optimizeDeps.include = depsConfig.optimizeDeps.include.filter(
+			(dep) => !isDisabled(dep)
+		);
+	}
```
bluwy, line 363: Currently doing this very cheeky handling for `disableDependencyReinclusion` which should still preserve it's original function.

dominikg (approved):

dominikg:
i'll try to add a testcase, afterwards its good to go

dominikg (changes requested):
added testcase is still failing in dev for a cjs dependency that is not optimized and during build cjs ends up in the output. build passes but browser fails.

packages/vite-plugin-svelte/src/utils/options.ts
```diff
@@ -361,103 +380,75 @@ export function buildExtraViteConfig(
...
-	const isExcluded = (dep: string) =>
-		isOptimizeExcluded(dep, exclude) || isOptimizeExcluded(dep, optimizeDeps?.exclude);
-	if (!isExcluded('svelte')) {
+	if (!isDepExcluded('svelte', config.optimizeDeps?.exclude ?? [])) {
 		const svelteImportsToInclude = SVELTE_IMPORTS.filter((x) => x !== 'svelte/ssr'); // not used on clientside
 		log.debug(
 			`adding bare svelte packages to optimizeDeps.include: ${svelteImportsToInclude.join(', ')} `
 		);
-		include.push(...svelteImportsToInclude.filter((x) => !isIncluded(x)));
+		include.push(...svelteImportsToInclude);
```
bluwy, line 436: note: I removed this filter as it's fine to include duplicate identifiers

dominikg (approved):

PR #479, opened 2022-10-28
feat: use Vite 3.2 `preprocessCSS` API for `useVitePreprocess`

bluwy:
And fallback to the Vite plugin hack if lower than 3.2


packages/vite-plugin-svelte/src/utils/preprocess.ts
```diff
@@ -1,14 +1,7 @@
-import {
-	transformWithEsbuild,
-	ESBuildOptions,
-	ResolvedConfig,
-	TransformResult,
-	Plugin
-} from 'vite';
+import { transformWithEsbuild, ESBuildOptions, ResolvedConfig, Plugin } from 'vite';
```
dominikg, line 1: can we use  `import * as vite from 'vite'` here and then skip the dynamic import below?
And maybe be specific about type imports.

bluwy: Good idea, didn't thought of that

bluwy: updated in f83c46c

dominikg (approved):

packages/vite-plugin-svelte/src/utils/preprocess.ts
```diff
@@ -40,34 +34,46 @@ function createViteScriptPreprocessor(): Preprocessor {
...
+// eslint-disable-next-line no-unused-vars
+type CssTransform = (code: string, filename: string) => Promise<{ code: string; map?: any }>;
+
+function getCssTransformFn(config: ResolvedConfig): CssTransform {
+	// API is only available in Vite 3.2 and above
+	if (vite.preprocessCSS) {
+		return async (code, filename) => {
+			return vite.preprocessCSS(code, filename, config);
+		};
+	} else {
```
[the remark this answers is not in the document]
bluwy: added in 94961d1

PR #482, opened 2022-11-03
fix: don't resolve via svelte field for prebundled libraries

dominikg:
so they actually get used

This pushes the svelte dependencies to optimizeDeps.include when prebundleSvelteLibraries is true so needs a bit more testing.

One sideeffect is that cold start of vite will be slower, but less restarts happen due to optimization being done earlier.


bluwy (approved):

benmccann:
Wouldn't this be roughly equivalent to https://github.com/sveltejs/vite-plugin-svelte/pull/459 for SvelteKit users?

dominikg:
No, this just prevents resolving dependencies via svelte field again after they have been prebundled. Which is a bug causing the prebundled dependencies not being used.

packages/vite-plugin-svelte/src/index.ts
```diff
@@ -154,21 +154,29 @@ export function svelte(inlineOptions?: Partial<Options>): Plugin[] {
...
-					if (resolved) {
-						log.debug(
-							`resolveId resolved ${resolved} via package.json svelte field of ${importee}`
+				//@ts-expect-error scan
+				const scan = !!opts?.scan; // scanner phase of optimizeDeps
+				const isPrebundled =
+					options.prebundleSvelteLibraries && viteConfig.optimizeDeps?.include?.includes(importee);
+				// for prebundled libraries we let vite resolve the prebundling result
+				// for ssr, during scanning and non-prebundled, we do it
+				if (ssr || scan || !isPrebundled) {
```
benmccann, line 163: If there is an `exports` then you potentially may resolve to a very different place during the scan and non-scan phases, which doesn't seem right to me?

packages/vite-plugin-svelte/src/index.ts
```diff
@@ -154,21 +154,29 @@ export function svelte(inlineOptions?: Partial<Options>): Plugin[] {
...
 				}
-				try {
-					const resolved = resolveViaPackageJsonSvelte(importee, importer, cache);
-					if (resolved) {
-						log.debug(
-							`resolveId resolved ${resolved} via package.json svelte field of ${importee}`
+				//@ts-expect-error scan
+				const scan = !!opts?.scan; // scanner phase of optimizeDeps
+				const isPrebundled =
+					options.prebundleSvelteLibraries && viteConfig.optimizeDeps?.include?.includes(importee);
```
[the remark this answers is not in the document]
dominikg: yes,` options.prebundleSvelteLibraries && !isExcluded` is the better check here. :+1:

packages/vite-plugin-svelte/src/index.ts
dominikg: no, thats excactly what we want. scan should find the package as described by the svelte field, optimize it and then it lives in `.vite/deps/something.js`  so after scan we want vite itself to resolve to that prebundled version.

packages/vite-plugin-svelte/src/index.ts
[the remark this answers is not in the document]
dominikg: done

packages/vite-plugin-svelte/src/index.ts
benmccann: Imagine the Svelte field points to "index.svelte" and the `exports` map has nothing but `package.json`. Then I think Vite would fail to resolve anything after scan?

dominikg: if the scan returned the .svelte file, the optimizer should have created a bundle in /node_modules/.vite for it. but now that i write this, thats not used during build. So we would additionally have to do it on build too unless optimize for build is set?

bluwy: Ah yeah we should probably handle build differently by checking `optimizeDeps.disabled` too. By default, build should always resolve `pkg.svelte` regardless of prebundling.

dominikg: done

bluwy (approved):

PR #483, opened 2022-11-03
chore: remove internal dependency type

benmccann:
We want to deprecate the `svelte` field, so remove reliance on it when crawling dependencies. Crawling more dependencies may have some small negative performance implications, but overall removing the `svelte` field will greatly improve performance by allowing us to more often resolve prebundled Svelte libraries which we can't do when the `svelte` field is present


benmccann:
This has been fixed in `main`, so this PR is no longer necessary

PR #495, opened 2022-11-12
fix: svelte version detection was broken

dominikg:
#490 introduced svelte version testing that didn't work as desired. inverted the condition and also updated the value used by esbuild plugin to inject css.

### DIFF
--- a/packages/vite-plugin-svelte/src/utils/esbuild.ts
+++ b/packages/vite-plugin-svelte/src/utils/esbuild.ts
@@ -43,7 +43,7 @@ async function compileSvelte(
 ): Promise<string> {
 	let css = options.compilerOptions.css;
 	if (css !== 'none') {
-		css = isCssString ? 'inject' : true;
+		css = isCssString ? 'injected' : true;
 	}
 	const compileOptions: CompileOptions = {
 		...options.compilerOptions,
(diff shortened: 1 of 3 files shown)


packages/vite-plugin-svelte/src/utils/esbuild.ts
```diff
@@ -43,7 +43,7 @@ async function compileSvelte(
 ): Promise<string> {
 	let css = options.compilerOptions.css;
 	if (css !== 'none') {
-		css = isCssString ? 'inject' : true;
+		css = isCssString ? 'injected' : true;
```
bluwy, line 46: I missed this too 😅 Didn't expect it to be past-tense.

packages/vite-plugin-svelte/src/utils/svelte-version.ts
```diff
@@ -32,5 +32,6 @@ export function compareToSvelte(version: string): 1 | 0 | -1 {
 }
 
 export function atLeastSvelte(version: string) {
-	return compareToSvelte(version) >= 0;
+	const result = compareToSvelte(version) <= 0;
+	return result;
```
bluwy, line 36: took me a while to grasp why but makes sense now. nice catch

bluwy (approved):

packages/vite-plugin-svelte/src/utils/svelte-version.ts
dominikg: still mad at myself for this one

PR #496, opened 2022-11-12
fix: no longer read kit.browser.hydrate

dominikg:
to initialize compilerOptiona.hydratable

### DIFF
--- a/packages/vite-plugin-svelte/src/utils/options.ts
+++ b/packages/vite-plugin-svelte/src/utils/options.ts
@@ -286,20 +286,9 @@ function removeIgnoredOptions(options: ResolvedOptions) {
 // some SvelteKit options need compilerOptions to work, so set them here.
 function addSvelteKitOptions(options: ResolvedOptions) {
 	// @ts-expect-error kit is not typed to avoid dependency on sveltekit
-	if (options?.kit != null) {
-		// @ts-expect-error kit is not typed to avoid dependency on sveltekit
-		const kit_browser_hydrate = options.kit.browser?.hydrate;
-		const hydratable = kit_browser_hydrate !== false;
-		if (
-			options.compilerOptions.hydratable != null &&
-			options.compilerOptions.hydratable !== hydratable
-		) {
-			log.warn(
-				`Conflicting values "compilerOptions.hydratable: ${options.compilerOptions.hydratable}" and "kit.browser.hydrate: ${kit_browser_hydrate}" in your svelte config. You should remove "compilerOptions.hydratable".`
-			);
-		}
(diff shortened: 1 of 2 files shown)


packages/vite-plugin-svelte/src/utils/options.ts
```diff
@@ -286,20 +286,9 @@ function removeIgnoredOptions(options: ResolvedOptions) {
...
-		) {
-			log.warn(
-				`Conflicting values "compilerOptions.hydratable: ${options.compilerOptions.hydratable}" and "kit.browser.hydrate: ${kit_browser_hydrate}" in your svelte config. You should remove "compilerOptions.hydratable".`
-			);
-		}
-		log.debug(`Setting compilerOptions.hydratable: ${hydratable} for SvelteKit`);
-		options.compilerOptions.hydratable = hydratable;
+	if (options?.kit != null && options.compilerOptions.hydratable == null) {
+		log.debug(`Setting compilerOptions.hydratable = true for SvelteKit`);
+		options.compilerOptions.hydratable = true;
```
bluwy, line 291: Shouldn't SvelteKit pass this as an inline config instead?

dominikg: then the user won't be able to set it to false in config file

bluwy: This feels a bit like the `prebundleSvelteLibraries` issue. I think we should have SvelteKit export their plugin as async instead so they can read the svelte config themselves and we can avoid these hard-coded patches.

dominikg: yes, ideally we find a way kit can control this in a way that doesn't prevent user overriding it themselves. 
It does already read svelte config on it's own and uses compilerOptions.hydratable too, see https://github.com/sveltejs/kit/pull/5155

dominikg: the main "problem" here is that svelte defaults hydratable to false whereas kit needs it to be true by default and currently kit doesn't have a nice way of supplying default values to compilerOptions

bluwy: It could leverage https://github.com/vitejs/vite/discussions/8509 though, that was one of the main goals but I didn't make the change in SvelteKit as it's technically a breaking change, but seeing the recent mini breaking changes maybe this is fine too? 😬

bluwy: Alright. Let's fix this for now and follow-up with a removal.

bluwy (approved):

PR #499, opened 2022-11-14
fix: bump svelte-hmr to support compileOptions.css='external'

dominikg:
the change in svelte-hmr is tiny and just for supporting css='external' properly. 
bump it just to avoid someone updating v-p-s without updating svelte-hmr

see https://github.com/sveltejs/svelte-hmr/pull/63


bluwy:
Looks like renovate covered this https://github.com/sveltejs/vite-plugin-svelte/pull/497

dominikg:
well, no changeset then

PR #502, opened 2022-11-18
feat: enable support for exports condition

dominikg:
This PR enables using an export condition `svelte` to point at svelte source files.

eg this exports map
```js
	"exports": {
		"./*": {
			"svelte": "./src/components/*"
		},
		".": {
			"svelte": "./index.js"
		}
	}
```
allows you to either `import {SomeThing} from 'some-lib'` when index.js exports SomeThing, or `import SomeThing from 'some-lib/SomeThing.svelte'`  if you prefer a deep import to avoid the index.

This is a very simplistic example and the real usecase is to still allow exporting compiled svelte files or other js files via the import condition.

See https://nodejs.org/api/packages.html#conditional-exports for what you can do with export conditions.

### DIFF
--- a/packages/vite-plugin-svelte/src/utils/options.ts
+++ b/packages/vite-plugin-svelte/src/utils/options.ts
@@ -398,7 +398,7 @@ async function buildExtraConfigForDependencies(options: PreResolvedOptions, conf
 			if (typeof pkgJson.exports === 'object') {
 				// use replacer as a simple way to iterate over nested keys
 				JSON.stringify(pkgJson.exports, (key, value) => {
-					if (key === 'svelte') {
+					if (SVELTE_EXPORT_CONDITIONS.includes(key)) {
 						hasSvelteCondition = true;
 					}
 					return value;
--- a/packages/vite-plugin-svelte/src/utils/options.ts
+++ b/packages/vite-plugin-svelte/src/utils/options.ts
@@ -2,7 +2,12 @@
 import { ConfigEnv, ResolvedConfig, UserConfig, ViteDevServer, normalizePath } from 'vite';
 import { log } from './log';
 import { loadSvelteConfig } from './load-svelte-config';
-import { SVELTE_HMR_IMPORTS, SVELTE_IMPORTS, SVELTE_RESOLVE_MAIN_FIELDS } from './constants';
+import {
+	SVELTE_EXPORT_CONDITIONS,
+	SVELTE_HMR_IMPORTS,
+	SVELTE_IMPORTS,
+	SVELTE_RESOLVE_MAIN_FIELDS
+} from './constants';
 // eslint-disable-next-line node/no-missing-import
 import type { CompileOptions, Warning } from 'svelte/types/compiler/interfaces';
 import type {
@@ -315,7 +320,8 @@ export async function buildExtraViteConfig(
 	const extraViteConfig: Partial<UserConfig> = {
 		resolve: {
 			mainFields: [...SVELTE_RESOLVE_MAIN_FIELDS],
-			dedupe: [...SVELTE_IMPORTS, ...SVELTE_HMR_IMPORTS]
+			dedupe: [...SVELTE_IMPORTS, ...SVELTE_HMR_IMPORTS],
+			conditions: [...SVELTE_EXPORT_CONDITIONS]
 		}
 		// this option is still awaiting a PR in vite to be supported
 		// see https://github.com/sveltejs/vite-plugin-svelte/issues/60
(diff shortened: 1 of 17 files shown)


benmccann (approved):

dominikg:
do we need additional tests, eg a scenario with mixes svelte field and exports condition? Is the usage i described above what we want to go for?
 i tried using the .svelte extension with glob but it didn't work even if the node docs show tha *.js is a thing.
 
 We only ever expect libraries to use the `.svelte` extension with preprocessed .svelte output, never other extensions like .svx that require additional markup preprocessors to be present, correct?
 
```js
	"exports": {
	       // boohoo, didn't work in my tests
		"./*.svelte": {
			"svelte": "./src/components/*.svelte"
		},
		".": {
			"svelte": "./index.js"
		}
	}
```

packages/vite-plugin-svelte/src/utils/options.ts
```diff
@@ -315,7 +320,8 @@ export async function buildExtraViteConfig(
 	const extraViteConfig: Partial<UserConfig> = {
 		resolve: {
 			mainFields: [...SVELTE_RESOLVE_MAIN_FIELDS],
-			dedupe: [...SVELTE_IMPORTS, ...SVELTE_HMR_IMPORTS]
+			dedupe: [...SVELTE_IMPORTS, ...SVELTE_HMR_IMPORTS],
+			conditions: [...SVELTE_EXPORT_CONDITIONS]
```
dominikg, line 324: we only add these conditions and let vite merge them. if there are other conditions already, svelte might have a lower priority. should we unshift them to try and ensure they are first?

benmccann: Yeah, I think svelte should come first

benmccann:
I wonder if it's just `*.svelte` that doesn't work or if other extensions are affected. My guess is that it's an existing Vite bug, but it might be nice to confirm and file an issue.

bluwy:
> i tried using the .svelte extension with glob but it didn't work even if the node docs show tha *.js is a thing.

Yeah I think this should technically work too, hoping it's only a bug on not some hidden nodejs spec.

packages/vite-plugin-svelte/src/utils/options.ts
dominikg: actually it looks like vite does not expose the default resolve extensions config, so the values we see here would all be custom conditions added by users or other plugins. i don't think we need to try and stay first of the custom ones.

Given that the testcase works i'm inclined to leave it that way until someone comes up with a strange nested condition need. @bluwy how does this work in astro if you use multiple frameworks at the same time that all require custom extensions?

bluwy: Oh yeah [looks like so](https://github.com/builderio/vite/blob/b80f7c744f3b14e725799c2a4ef69af97d738216/packages/vite/src/node/config.ts#L485). So we could leave it as is 👍 



> @bluwy how does this work in astro if you use multiple frameworks at the same time that all require custom extensions?

I don't think it'll be a problem as it doesn't make sense to do `pkg.exports['./'].astro.svelte` or `pkg.exports['./'].svelte.solid`.

packages/vite-plugin-svelte/src/utils/options.ts
```diff
@@ -363,6 +369,7 @@ export async function buildExtraViteConfig(
 		// Currently a placeholder as more information is needed after Vite config is resolved,
 		// the real Svelte plugin is added in `patchResolvedViteConfig()`
 		extraViteConfig.optimizeDeps.esbuildOptions = {
+			conditions: [...SVELTE_EXPORT_CONDITIONS],
```
bluwy, line 372: I don't think this is needed as Vite should handle the conditions too. Vite doesn't set the conditions as an esbuild option, but all path resolve will go through Vite's custom resolver so it's safe too.

bluwy (approved):
one nit but lgtm!

packages/vite-plugin-svelte/src/utils/options.ts
dominikg: it didn't work without iirc but can test again. maybe vite itself should propagate custom conditions to esbuild from resolve.conditions?

bluwy: that would also make sense as a safety measure in Vite, but theoretically it shouldn't need to 🤔

packages/vite-plugin-svelte/src/utils/options.ts
dominikg: thanks. so we'll leave it as is for now

PR #503, opened 2022-11-20
feat: log compile stats

dominikg:
We want to set `prebundleSvelteLibraries: true` by default and this has implications for users.

Esp. with large component libraries it can lead to significant delays when loading the dev page in the browser or building.
To make these transparent, we collect compile time statistics and log them.

example output for dev:
```  VITE v3.2.3  ready in 541 ms

  ➜  Local:   http://localhost:5173/
5:43:05 PM [vite-plugin-svelte] prebundle libraries in progress ...
5:43:19 PM [vite-plugin-svelte] ssr compile in progress ...
5:43:24 PM [vite-plugin-svelte] prebundle libraries done.
package                         files     time     avg
carbon-icons-svelte              2002   10.02s   5.0ms
carbon-components-svelte          206    2.42s  11.7ms
5:43:28 PM [vite-plugin-svelte] ssr compile done.
package                                 files    time     avg
carbon-icons-svelte                      2002   2.68s   1.3ms
carbon-components-svelte                  206   0.53s   2.6ms
playground-big-component-library-kit        3   9.0ms   3.0ms
```

example output for build:
```
> vite build

vite v3.2.3 building for production...
✓ 2333 modules transformed.
5:44:47 PM [vite-plugin-svelte] dom compile done.
package                                 files     time     avg
carbon-icons-svelte                      2002    9.02s   4.5ms
carbon-components-svelte                  206    2.27s  11.0ms
playground-big-component-library-kit        4   84.6ms  21.1ms
@sveltejs/kit                               1   22.0ms  22.0ms
vite v3.2.3 building SSR bundle for production...
✓ 2351 modules transformed.
5:44:55 PM [vite-plugin-svelte] ssr compile done.
package                                 files    time     avg
carbon-icons-svelte                      2002   2.96s   1.5ms
carbon-components-svelte                  206   0.59s   2.9ms
playground-big-component-library-kit        4   8.6ms   2.1ms
@sveltejs/kit                               1   1.9ms   1.9ms
```

These stats are logged at the `info` level.

You can disable them via the new `experimental.disableCompileStats` option for vite-plugin-svelte
```
// svelte.config.js
export default {
  vitePlugin: {
    experimental: {
      disableCompileStats: true, // true disables completely, 'dev' for dev and 'build' for build
    }
  }
}
```


packages/vite-plugin-svelte/src/utils/compile.ts
```diff
@@ -18,6 +20,31 @@ const _createCompileSvelte = (makeHot: Function) =>
...
+					stats = options.stats.startCollection(`${ssr ? 'ssr' : 'dom'} compile`, {
+						logInProgress: () => false
+					});
+				}
+			} else {
+				// dev time ssr, it's a ssr request and there are no stats, assume new page load and start collecting
+				if (ssr && !stats) {
+					stats = options.stats.startCollection('ssr compile');
+				}
+				// stats are being collected but this isn't an ssr request, assume page loaded and stop collecting
```
benmccann, line 37: will this cause anything to print if ssr is disabled? (probably we won't want it to in that case)

I wonder if you could use a hook like `writeBundle` to understand when a request was done being processed?

packages/vite-plugin-svelte/src/utils/vite-plugin-svelte-stats.ts
```diff
@@ -0,0 +1,208 @@
...
+	private async _finish(collection: StatCollection) {
+		collection.finished = true;
+		const now = performance.now();
+		collection.duration = now - collection.collectionStart;
+		const logResult = collection.options.logResult(collection);
+		if (logResult) {
+			await this._aggregateStatsResult(collection);
+			log.info(`${collection.name} done.`, formatPackageStats(collection.packageStats!));
+		}
+		//cut some ties to free it for garbage collection
```
benmccann, line 143: the other comments have a space after `//`
```suggestion
		// cut some ties to free it for garbage collection
```

packages/vite-plugin-svelte/src/utils/compile.ts
dominikg: this is for dev, so no writeBundle calls.  During dev we only log ssr stats because thats what we are able to batch up reliably. 
In theory it is possible to use a debounce to call finish on the dom compile but this can get very messy too. As the dom requests visible in the browser network log relate to 1 compile call (each import), these stats would be nice to have, but are less essential than the ssr stats that are currently completely hidden

dominikg:
added an option to disable because `--logLevel warn` doesn't seem to propagate to the second vite build kit kicks off and it also removes other useful logging.

packages/vite-plugin-svelte/src/utils/vite-plugin-svelte-stats.ts
```diff
@@ -0,0 +1,206 @@
...
+					let path = pkgPath?.replace(/package.json$/, '');
+					let name = JSON.parse(readFileSync(pkgPath, 'utf-8')).name;
+					if (!name) {
+						// some packages have nameless nested package.json
+						pkgPath = await findClosestPkgJsonPath(path);
+						if (pkgPath) {
+							path = pkgPath?.replace(/package.json$/, '');
+							name = JSON.parse(readFileSync(pkgPath, 'utf-8')).name;
+						}
+					}
```
bluwy, line 174: This is.. interesting 😄 I guess it would be nice if vitefu exported a more flexible API for this, but I don't usually mind code like this for now.

docs/config.md
```diff
@@ -208,6 +208,13 @@ A [picomatch pattern](https://github.com/micromatch/picomatch), or array of patt
 
   Force Vite to pre-bundle Svelte libraries. Setting this `true` should improve initial page load performance, especially when using large Svelte libraries. See the [FAQ](./faq.md#what-is-going-on-with-vite-and-pre-bundling-dependencies) for details of the pre-bundling implementation.
 
+### disableCompileStats
+
+- **Type** `boolean | 'dev' | 'build'`
+- **Default:** `false`
+
+  disable svelte compile statistics
```
bluwy, line 216: > added an option to disable because --logLevel warn doesn't seem to propagate to the second vite build kit kicks off and it also removes other useful logging.

I feel like we should fix this in SvelteKit instead another option. It would be nice to reduce the number of options, similar to Vite's idea.

I also wonder if the logs should be enabled via `DEBUG=`, but we could also not do that as a start until someone finds it annoying 🤔

dominikg: fixing logLevel passthrough in kit (or vite) is indeed a good idea. But being able to disable the stats logs separately still makes sense. 
The debug log is an interesting proposition, but we'd have to restructure how we debug log today and segregate that into extra categories so that you could do `DEBUG=vite-plugin-svelte:stats:*` or something to not overload the output.

I want the logs to be visible by default so that when we enable prebundleSvelteLibraries by default, users don't get surprised by the initial wait that might cause. That isn't possible with debug i think

packages/vite-plugin-svelte/src/utils/vite-plugin-svelte-stats.ts
dominikg: yeah, not proud of that one, but it works :neutral_face:

packages/vite-plugin-svelte/src/utils/compile.ts
benmccann: it makes me wonder if we should have a feature request to Vite to add a request ID. but I wouldn't let that block this

docs/config.md
benmccann: my feeling for now is that we should have them enabled by default. maybe I'll change my mind after experiencing it a bit more on real projects, but they seem to be pretty helpful

I'm torn on the option to disable them. I can see that being valuable, but it also possibly means we can't remove the option without a breaking change, so I somewhat wonder if we should wait until someone asks for it. But at the same time it's already implemented :shrug:

dominikg: we can move the disable option to experimental :see_no_evil:

dominikg: done.

packages/vite-plugin-svelte/src/utils/compile.ts
dominikg: interesting idea, though i wonder how that works, esp when you do clientside nav in kit. But we can move that off this comment thread.

bluwy (approved):
Moving the option to `experimental` sounds fine to me 👍 I hope we can start reducing the options for v2 though so there's less to document and nicer ootb experience.

PR #507, opened 2022-11-23
fix: use a recursive helper to find named parent package

dominikg:
catch errors to avoid breaking the process just because stats didn't work

### DIFF
--- a/packages/vite-plugin-svelte/src/utils/vite-plugin-svelte-stats.ts
+++ b/packages/vite-plugin-svelte/src/utils/vite-plugin-svelte-stats.ts
@@ -2,6 +2,7 @@ import { log } from './log';
 //eslint-disable-next-line node/no-missing-import
 import { findClosestPkgJsonPath } from 'vitefu';
 import { readFileSync } from 'fs';
+import { dirname } from 'path';
 import { performance } from 'perf_hooks';
 
 interface Stat {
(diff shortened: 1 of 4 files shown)


packages/vite-plugin-svelte/src/utils/vite-plugin-svelte-stats.ts
```diff
@@ -85,6 +86,31 @@ function formatPackageStats(pkgStats: PackageStats[]) {
...
+	if (pkgPath) {
+		const name = JSON.parse(readFileSync(pkgPath, 'utf-8')).name;
+		if (!name) {
+			return parseClosestNamedPackage(file, dirname(pkgPath));
+		}
+		const path = pkgPath.replace(/package.json$/, '');
+		return { name, path };
+	} else {
+		// no package.json found with a name in it, give up.
+		const path = dirname(file) + '/';
```
bluwy, line 103: maybe we should use `path.join` to be safe. needs to update the import though.
```suggestion
		const path = path.join(dirname(file), '/');
```

dominikg: it was worse, vitefu doesn't normalize so cached paths were not found by the `file.startsWith(path)` check. Updated to include normalization

bluwy (approved):

bluwy (approved):

PR #509, opened 2022-11-24
Export `vitePreprocess` API

bluwy:
Export `vitePreprocess()` Svelte preprocessor from `@sveltejs/vite-plugin-svelte`. (Previously exported from `/preprocess` subpath but now root)

`experimental.useVitePreprocess` is deprecated.

---

Things I'm still thinking of, but could be a future problem:
1. Should we deprecate `useVitePreprocess`?
2. Following no1, there are frameworks like Astro that wants to preprocess with Vite by default. The user can get intellisense by adding `vitePreprocess` to `svelte.config.js` but that results in two same preprocessors running at the same time. Should we deduplicate them internally?
3. When running in a Vite dev server that has it's own `ResolvedConfig`, do we want to share it to the style preprocessor? If the user passes an `InlineConfig` to the style preprocessor, it makes config merging not possible. Also it feels a bit magical 🤔 

EDIT: Answered:
1. yes.
2. Astro should deduplicate themselves.
3. now implemented in this PR. When using the preprocessor in Vite dev server, the preprocessor will use that config.


dominikg:
yes to deprecate, the improvement in DX makes using the plugin option inferior. we can add docs how to migrate. it's experimental still so no semver constraints

docs/preprocess.md
```diff
@@ -0,0 +1,30 @@
+# Preprocess
+
+`@sveltejs/vite-plugin-svelte/preprocess` exports utilities to preprocess Svelte components using Vite's built-in transformers.
+
+## vitePreprocess
+
+- **Type:** `{ script?: boolean, style?: boolean | InlineConfig | ResolvedConfig }`
+- **Default:** `{ script: true, style: true }`
+
+  A Svelte preprocessor that supports transforming TypeScript, PostCSS, SCSS, Less, Stylus, and SugarSS. These are transformed when the script or style tags have the respective `lang` attribute.
```
benmccann, line 10:
```suggestion
A Svelte preprocessor that supports transforming TypeScript, PostCSS, SCSS, Less, Stylus, and SugarSS. These are transformed when the script or style tags have the respective `lang` attribute.
```

docs/preprocess.md
```diff
@@ -0,0 +1,30 @@
...
+
+  - TypeScript: `<script lang="ts">`
+  - PostCSS: `<style lang="postcss">`
+  - SCSS: `<style lang="scss">`
+  - Less: `<style lang="less">`
+  - Stylus: `<style lang="stylus">`
+  - SugarSS: `<style lang="sss">`
+
+  If required, you can turn script or style transforming off by setting the `script` or `style` option to `false`. The `style` option also accepts Vite's `InlineConfig` and `ResolvedConfig` types for advanced usage.
```
benmccann, line 19:
```suggestion
If required, you can turn script or style transforming off by setting the `script` or `style` option to `false`. The `style` option also accepts Vite's `InlineConfig` and `ResolvedConfig` types for advanced usage.
```

docs/preprocess.md
```diff
@@ -0,0 +1,30 @@
+# Preprocess
```
benmccann, line 1: It would be good to document what the advantages / disadvantages are of using this feature vs `svelte-preprocess`

dominikg: 2 largest selling points are ts transform with esbuild and style url resolve. followed by general niceties of using the same css pipeline as vite, so less configuration.

docs/preprocess.md
bluwy: The indent is intentional with the `config.md` convention

docs/preprocess.md
bluwy: Good idea. `url()` isn't actually resolved in preprocessing as Vite will do another pass when it becomes a virtual module and resolve from there. This does make less configuration though so would be good to document.

docs/preprocess.md
```diff
@@ -0,0 +1,32 @@
...
+
+`@sveltejs/vite-plugin-svelte/preprocess` exports utilities to preprocess Svelte components using Vite's built-in transformers.
+
+Compared to [`svelte-preprocess`](https://github.com/sveltejs/svelte-preprocess), Vite preprocessors shares the same CSS configuration from the Vite config so you don't have to configure them twice. [`esbuild`](http://esbuild.github.io) is also used to transform TypeScript by default.
```
benmccann, line 5:
```suggestion
Compared to [`svelte-preprocess`](https://github.com/sveltejs/svelte-preprocess), Vite preprocessors share the same CSS configuration from the Vite config so you don't have to configure them twice. [`esbuild`](http://esbuild.github.io) is also used to transform TypeScript by default.
```

benmccann:
I'd be curious if there's been any rough benchmarking to see if this is faster. esbuild is obviously faster, but you'd also have to call out to another process which is slower (I think once per file), so I wonder if it's overall faster. If we can confirm it's faster, I wonder if this should be made the default in `create-svelte`?

docs/preprocess.md
```diff
@@ -0,0 +1,34 @@
...
+  - Stylus: `<style lang="stylus">`
+  - SugarSS: `<style lang="sss">`
+
+  If required, you can turn script or style transforming off by setting the `script` or `style` option to `false`. The `style` option also accepts Vite's `InlineConfig` and `ResolvedConfig` types for advanced usage.
+
+  **Example:**
+
+  ```js
+  // svelte.config.js
+  import { vitePreprocess } from '@sveltejs/vite-plugin-svelte/preprocess';
```
dominikg, line 29:
```suggestion
  import { vitePreprocess } from '@sveltejs/vite-plugin-svelte';
```

docs/preprocess.md
```diff
@@ -0,0 +1,34 @@
+# Preprocess
+
+`vite-plugin-svelte` also exports Vite preprocessors to preprocess Svelte components using Vite's built-in transformers.
+
+Compared to [`svelte-preprocess`](https://github.com/sveltejs/svelte-preprocess), Vite preprocessors share the same CSS configuration from the Vite config so you don't have to configure them twice. [`esbuild`](http://esbuild.github.io) is also used to transform TypeScript by default.
+
+However, `svelte-preprocess` does provide extra functionalities not available with Vite preprocessors, such as [template tag](https://github.com/sveltejs/svelte-preprocess#template-tag), [external files](https://github.com/sveltejs/svelte-preprocess#external-files), and [global styles](https://github.com/sveltejs/svelte-preprocess#global-style) ([It's recommended to use import instead](./faq.md#where-should-i-put-my-global-styles)). If those features are required, you can still use `svelte-preprocess`, but make sure to turn off it's script and style preprocessing options.
```
benmccann, line 7:
```suggestion
However, `svelte-preprocess` does provide extra functionalities not available with Vite preprocessors, such as [template tag](https://github.com/sveltejs/svelte-preprocess#template-tag), [external files](https://github.com/sveltejs/svelte-preprocess#external-files), and [global styles](https://github.com/sveltejs/svelte-preprocess#global-style) ([though it's recommended to use import instead](./faq.md#where-should-i-put-my-global-styles)). If those features are required, you can still use `svelte-preprocess`, but make sure to turn off it's script and style preprocessing options.
```

dominikg: should we mention all these extra features of svelte-preprocess at all? they are all not recommended to be used with vite. I'd rather not promote them

dominikg (approved):

PR #510, opened 2022-11-24
feat: log deprecation warning if svelte field causes difference in resolve

dominikg:
- [ ] add documentation and update log message to contain link

1) do we need an option to disable these warnings
2) should they be logged upon encountering them or batched in buildEnd
3) how do we test this

### DIFF
--- a/packages/vite-plugin-svelte/src/utils/vite-plugin-svelte-cache.ts
+++ b/packages/vite-plugin-svelte/src/utils/vite-plugin-svelte-cache.ts
@@ -1,5 +1,17 @@
 import { SvelteRequest } from './id';
 import { Code, CompileData } from './compile';
+import { readFileSync } from 'fs';
+import { dirname } from 'path';
+//eslint-disable-next-line node/no-missing-import
+import { findClosestPkgJsonPath } from 'vitefu';
+import { normalizePath } from 'vite';
+
+interface PackageInfo {
+	name: string;
+	version: string;
+	svelte?: string;
+	path: string;
+}
 
 export class VitePluginSvelteCache {
 	private _css = new Map<string, Code>();
@@ -8,6 +20,7 @@ export class VitePluginSvelteCache {
 	private _dependants = new Map<string, Set<string>>();
 	private _resolvedSvelteFields = new Map<string, string>();
 	private _errors = new Map<string, any>();
+	private _packageInfos: PackageInfo[] = [];
 
 	public update(compileData: CompileData) {
 		this._errors.delete(compileData.normalizedFilename);
@@ -124,4 +137,43 @@ export class VitePluginSvelteCache {
 	private _getResolvedSvelteFieldKey(importee: string, importer?: string): string {
 		return importer ? `${importer} > ${importee}` : importee;
 	}
+
+	public async getPackageInfo(file: string): Promise<PackageInfo> {
+		let info = this._packageInfos.find((pi) => file.startsWith(pi.path));
+		if (!info) {
+			info = await findPackageInfo(file);
+			this._packageInfos.push(info);
+		}
+		return info;
+	}
+}
+
+/**
+ * utility to get some info from the package.json a file belongs to (closest one with a "name" set)
+ *
+ * @param {string} file to find info for
+ * @returns {PackageInfo}
+ */
+async function findPackageInfo(file: string): Promise<PackageInfo> {
+	const info: PackageInfo = {
+		name: '$unknown',
+		version: '0.0.0-unknown',
+		path: '$unknown'
+	};
+	let path = await findClosestPkgJsonPath(file, (pkgPath) => {
+		const pkg = JSON.parse(readFileSync(pkgPath, 'utf-8'));
+		if (pkg.name != null) {
+			info.name = pkg.name;
+			if (pkg.version != null) {
+				info.version = pkg.version;
+			}
+			info.svelte = pkg.svelte;
+			return true;
+		}
+		return false;
+	});
+	// return normalized path with appended '/' so .startsWith works for future file checks
+	path = normalizePath(dirname(path ?? file)) + '/';
+	info.path = path;
+	return info;
 }
--- a/.changeset/cold-horses-begin.md
+++ b/.changeset/cold-horses-begin.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/vite-plugin-svelte': patch
+---
+
+log deprecation warnings for packages that use the "svelte" field to resolve svelte files differently than standard vite resolve
--- a/packages/vite-plugin-svelte/src/index.ts
+++ b/packages/vite-plugin-svelte/src/index.ts
@@ -76,7 +76,7 @@ export function svelte(inlineOptions?: Partial<Options>): Plugin[] {
 			},
 
 			async configResolved(config) {
-				options = resolveOptions(options, config);
+				options = resolveOptions(options, config, cache);
 				patchResolvedViteConfig(config, options);
 				requestParser = buildIdParser(options);
 				compileSvelte = createCompileSvelte(options);
@@ -169,9 +169,26 @@ export function svelte(inlineOptions?: Partial<Options>): Plugin[] {
 					try {
 						const resolved = await resolveViaPackageJsonSvelte(importee, importer, cache);
 						if (resolved) {
-							log.debug(
+							log.debug.once(
 								`resolveId resolved ${resolved} via package.json svelte field of ${importee}`
 							);
+							try {
+								const viteResolved = (
+									await this.resolve(importee, importer, { ...opts, skipSelf: true })
+								)?.id;
+								if (resolved !== viteResolved) {
+									const packageInfo = await cache.getPackageInfo(resolved);
+									log.warn.once(
+										`DEPRECATION WARNING: ${packageInfo.name}@${packageInfo.version} package.json has \`"svelte":"${packageInfo.svelte}"\` which resolves .svelte files differently than standard vite resolve.\nUsing the "svelte" field in package.json is deprecated and packages should use the "svelte" exports condition instead. See :LINK HERE: for more information.`
+									);
+								}
+							} catch (e) {
+								const packageInfo = await cache.getPackageInfo(resolved);
+								log.warn.once(
+									`DEPRECATION WARNING: ${packageInfo.name}@${packageInfo.version} package.json has \`"svelte":"${packageInfo.svelte}"\` which resolves .svelte files but standard vite resolve failed to resolve.\nUsing the "svelte" field in package.json is deprecated and packages should use the "svelte" exports condition instead. See :LINK HERE: for more information.`
+								);
+							}
+
 							return resolved;
 						}
 					} catch (e) {
--- a/packages/vite-plugin-svelte/src/utils/compile.ts
+++ b/packages/vite-plugin-svelte/src/utils/compile.ts
@@ -113,6 +113,7 @@ const _createCompileSvelte = (makeHot: Function) => {
 			compiled.js.code = makeHot({
 				id: filename,
 				compiledCode: compiled.js.code,
+				// @ts-expect-error hot isn't a boolean anymore at this point
 				hotOptions: { ...options.hot, injectCss: options.hot?.injectCss === true && hasCss },
 				compiled,
 				originalCode: code,
--- a/packages/vite-plugin-svelte/src/utils/log.ts
+++ b/packages/vite-plugin-svelte/src/utils/log.ts
@@ -73,7 +73,7 @@ function createLogger(level: string): LogFn {
 	const logFn: LogFn = _log.bind(null, logger) as LogFn;
 	const logged = new Set<String>();
 	const once = function (message: string, payload?: any) {
-		if (logged.has(message)) {
+		if (!logger.enabled || logged.has(message)) {
 			return;
 		}
 		logged.add(message);
--- a/packages/vite-plugin-svelte/src/utils/options.ts
+++ b/packages/vite-plugin-svelte/src/utils/options.ts
@@ -33,6 +33,7 @@ import {
 import { atLeastSvelte } from './svelte-version';
 import { isCommonDepWithoutSvelteField } from './dependencies';
 import { VitePluginSvelteStats } from './vite-plugin-svelte-stats';
+import { VitePluginSvelteCache } from './vite-plugin-svelte-cache';
 
 // svelte 3.53.0 changed compilerOptions.css from boolean to string | boolen, use string when available
 const cssAsString = atLeastSvelte('3.53.0');
@@ -180,7 +181,8 @@ function mergeConfigs<T>(...configs: (Partial<T> | undefined)[]): T {
 // also validates the final config.
 export function resolveOptions(
 	preResolveOptions: PreResolvedOptions,
-	viteConfig: ResolvedConfig
+	viteConfig: ResolvedConfig,
+	cache: VitePluginSvelteCache
 ): ResolvedOptions {
 	const css = cssAsString
 		? preResolveOptions.emitCss
@@ -217,7 +219,7 @@ export function resolveOptions(
 	const statsEnabled =
 		disableCompileStats !== true && disableCompileStats !== (merged.isBuild ? 'build' : 'dev');
 	if (statsEnabled && isLogLevelInfo) {
-		merged.stats = new VitePluginSvelteStats();
+		merged.stats = new VitePluginSvelteStats(cache);
 	}
 	return merged;
 }
--- a/packages/vite-plugin-svelte/src/utils/vite-plugin-svelte-stats.ts
+++ b/packages/vite-plugin-svelte/src/utils/vite-plugin-svelte-stats.ts
@@ -1,10 +1,7 @@
 import { log } from './log';
-//eslint-disable-next-line node/no-missing-import
-import { findClosestPkgJsonPath } from 'vitefu';
-import { readFileSync } from 'fs';
-import { dirname } from 'path';
 import { performance } from 'perf_hooks';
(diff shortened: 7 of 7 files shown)


dominikg:
example warning
```
11:46:09 AM [vite-plugin-svelte] DEPRECATION WARNING: e2e-test-dep-svelte-nested@1.0.0 package.json has `"svelte":"src/index.js"` which resolves .svelte files but standard vite resolve failed to resolve.
Using the "svelte" field in package.json is deprecated and packages should use the "svelte" exports condition instead. See :LINK HERE: for more information.
```

docs/faq.md
```diff
@@ -177,3 +177,32 @@ There is no golden rule, but you can follow these recommendations:
 
 This warning only occurs if you use non-default settings in your vite config that can cause problems in combination with prebundleSvelteLibraries.
 You should not use prebundleSvelteLibraries during build or for ssr, disable one of the incompatible options to make that warning (and subsequent errors) go away.
+
+<!-- the following header generates an anchor that is used in logging, do not modify!-->
+
+### deprecated "svelte" field
```
benmccann, line 183: per @dummdidumm's point, I'm not sure we need to actually deprecate the `svelte` field. we just need to stop resolving it before `exports` and resolve it after instead. that would make the warning trigger less frequently which would be nicer for users

benmccann: actually, I just looked at the code and I think it works assuming the svelte field will be left in place and only changes the order. if we really did want to get rid of it entirely we'd probably have to change the code, but I'd just suggest changing the title here and note below saying it will be removed and instead say the resolve order will change

dominikg: i don't think we can recover from a resolve error unless we _always_ call `this.resolve` in a try/catch  not something i want to do long term.

dominikg: also having 2 accepted ways to do this doesn't feel great. it makes writing new bundler plugins even more challanging. The only point i see in this whole excercise is that at the end of it the svelte field is no longer used at all.

docs/faq.md
```diff
@@ -177,3 +177,32 @@ There is no golden rule, but you can follow these recommendations:
...
+
+Example:
+
+```diff
+// package.json
+- "svelte": "src/index.js"
++ "exports": {
++   "./package.json": "./package.json",
++   "./*": {
++     "svelte": "./src/*",
```
Rich-Harris, line 197: does this need to be this...

```suggestion
+     "svelte": "./src/*.svelte",
```

...in order for `library/src/Foo.svelte` to be importable as `library/Foo`? Or is that undesirable/ (Would that pattern even work?)

dominikg: ideally it would be `"./*.svelte":{"svelte":"./src/*.svelte"}` but that didn't work in my testing even though node docs show a similar pattern for `*.js` 

omitting the extension altogether might confuse tools that look for it and may cause problems when there is a foo.js and a Foo.svelte  (hello windows)

docs/faq.md
benmccann: There's already two ways to resolve JS files with `main` and `exports`, so it's just the analogous thing. As long as we prefer `exports` we can rip out all the code in `vite-plugin-svelte` and just defer to Vite

dominikg: how do you expect
> **resolve it after instead.**

to happen after

> rip out all the code in vite-plugin-svelte and just defer to Vite

?

right now vite-plugin-svelte is a `pre` plugin exactly for the reason that we need to resolve via svelte field before vite resolve has a chance to error. Resolving the svelte field **after** a vite resolve error isn't possible unless you do in a pre-plugin. If you need to keep support for that alive after vite-plugin-svelte has switched it off, it should live in a legacy plugin that only users install that still have libraries in their project that would break otherwise. i don't want to burden regular use with that.

```js
// shortend 
async resolveId(importee,importer,options) {
  try {
    return this.resolve(importee,importer,{...options,skipSelf: true});
  } catch (e) {
    // vite resolve failed for any reason,
   const svelteResolved = await resolveViaPackageJsonSvelteField(importee,importer,options);
   if(svelteResolved) {
      log.warn('some warning here')
      return svelteResolved;
   }
   throw e;
  }
}

Comment (benmccann) [reply docs/faq.md:183]: `vite-plugin-svelte` wouldn't resolve it at all. We've set `svelte` in [`resolve.mainFields`](https://vitejs.dev/config/shared-options.html#resolve-mainfields), so Vite will handle it.

Comment (bluwy) [reply docs/faq.md:183]: I think we should phase out `pkg.svelte` too. `pkg.exports` is the official way to add custom condition names. Adding custom top-level fields isn't a standard and was only chosen because `pkg.exports` didn't exist then.

File: docs/faq.md (hunk 2 above)
Comment (bluwy) [reply docs/faq.md:197]: I think Rich's suggestion is the correct way to support it. Using `"./*.svelte"` means they have to import as `my-lib/Foo.svelte`

File: docs/faq.md [hunk 3]
```diff
@@ -177,3 +177,32 @@ There is no golden rule, but you can follow these recommendations:
...
+This field requires a custom implementation to resolve, so you have to use a bundler plugin and this plugin needs to implement resolving.
+Since then, node has added support for [conditional exports](https://nodejs.org/api/packages.html#conditional-exports), which have more generic support in bundlers and node itself. So to increase the compatibility with the wider ecosystem and reduce the implementation needs for current and future bundler plugins, it is recommended that packages use the "svelte" exports condition.
+
+Example:
+
+```diff
+// package.json
+- "svelte": "src/index.js"
++ "exports": {
++   "./package.json": "./package.json",
```
bluwy, line 195: Maybe we should not recommend export `package.json` too as it's not necessarily needed.

dominikg: this is what rollup plugin svelte did recommend and before we moved it to vitefu vite-plugin-svelte still used that export to load package.json if possible. does vitefu not do it and always read from fs?

docs/faq.md
dominikg: i'm not an expert on building sane exports map for a new custom condition, so definetely up for improvements. Unfortunately i couldn't make the case with .svelte in the condition work, would love for someone to show a working example.

docs/faq.md
dominikg: > `vite-plugin-svelte` wouldn't resolve it at all. We've set `svelte` in [`resolve.mainFields`](https://vitejs.dev/config/shared-options.html#resolve-mainfields), so Vite will handle it.

no it won't. if exports entry exists, mainFields are ignored. mainFields also don't resolve non-js extensions afaik. There is no way to "fall back to svelte field" without wrapping vite resolve in a try catch - at least that i know of.

docs/faq.md
bluwy: yeah vitefu will always read from the fs. I don't think there's perf issues in practice though. There's https://github.com/sveltejs/rollup-plugin-svelte/pull/198 in rollup plugin svelte to stop warning if `package.json` is not exported too.

benmccann: we should definitely have the plugin read from the fs. `rollup-plugin-svelte`'s behavior is incredibly annoying. that being said, I would leave this as is for now. `rollup-plugin-svelte` still will issue a warning if it's not exported, so users should probably export it for the time being until we update `rollup-plugin-svelte`

docs/faq.md
benmccann: > if exports entry exists, mainFields are ignored

I agree with that

> mainFields also don't resolve non-js extensions afaik

This would be news to me

benmccann: > mainFields also don't resolve non-js extensions afaik

I just tested this and it appears that Vite does resolve them. In order to test I installed a component (`svelte-simple-modal`), tweaked the `package.json` in the downloaded package to only have a `svelte` field pointing to a `.svelte` file, and then commented out `vite-plugin-svelte`'s call to `resolveViaPackageJsonSvelte`. It rendered the imported component just fine

docs/faq.md
benmccann: From https://github.com/nodejs/node/issues/46074#issuecomment-1370102411:

> As the creator of Rollup, I can empathise: allowing modules to import each other without specifying the .js extension was a silly mistake that added untold complexity for no good reason

If it's a bad idea for `.js` files, it's probably a bad idea for `.svelte` files too, right? I don't think I would change the example to make it work without the `.svelte` extension

dominikg:
I have updated the PR to latest main and refactored it to log at buildEnd.

The new log output looks like this:
```
11:03:00 PM [vite-plugin-svelte] WARNING: The following packages use a svelte resolve configuration in package.json that is going to cause problems in the future.

e2e-test-dep-svelte-exports-simple@1.0.0

Please see https://github.com/sveltejs/vite-plugin-svelte/blob/main/docs/faq.md#deprecated-svelte-field for details.
```

docs/faq.md
dominikg: removed the package.json line and also kept the svelte field as neutral instead of remove. We should not encourage them to remove it for now as that could break older tooling that still relies on it.

docs/faq.md
dominikg: So what does an ideal svelte exports condition look like today? just exporting the index.js that lists all .svelte files? Or use a glob too so that deep importing components can work?

I'd love to have a guideline for package authors here. Prebundling has made it less needed to use deep imports but they can still be a valid choice.

docs/faq.md
dominikg: @bluwy  iirc you changed the behavior of exports + mainfields lately. Is the above still true? If yes we should be able to remove the custom code for resolveViaPackageJsonSvelte even today.

docs/config.md
```diff
@@ -410,3 +410,10 @@ export default {
 - **Default:** `false`
 
   disable svelte compile statistics.
+
+### disableSvelteResolveWarnings
+
+- **Type** `boolean`
+- **Default:** `false`
+
+  disable svelte resolve warnings
```
benmccann, line 419: I don't know if we should make this option available. If we do, we should discourage its use
```suggestion
  disable svelte resolve warnings. Note: this is highly discouraged and you should instead fix these packages which will break in the future.
```

packages/e2e-tests/resolve-exports-svelte/vite.config.js
```diff
@@ -3,5 +3,10 @@ import { svelte } from '@sveltejs/vite-plugin-svelte';
 
 // https://vitejs.dev/config/
 export default defineConfig({
-	plugins: [svelte({ compilerOptions: { css: 'none' } })]
+	plugins: [
+		svelte({
+			compilerOptions: { css: 'none' },
+			experimental: { disableSvelteResolveWarnings: false }
```
benmccann, line 9: I think we can remove this line since it's false by default
```suggestion
			compilerOptions: { css: 'none' }
```

packages/vite-plugin-svelte/src/utils/vite-plugin-svelte-cache.ts
```diff
@@ -124,4 +140,43 @@ export class VitePluginSvelteCache {
...
+		if (!info) {
+			info = await findPackageInfo(file);
+			this._packageInfos.push(info);
+		}
+		return info;
+	}
+}
+
+/**
+ * utility to get some info from the package.json a file belongs to (closest one with a "name" set)
```
benmccann, line 155: could simplify this description:
```suggestion
 * utility to get some info from the closest package.json with a "name" set
```

packages/e2e-tests/resolve-exports-svelte/vite.config.js
dominikg: oops, this was from my local testing, keeping it there for easier switching wont hurt. its an e2e test app, so not affecting users in any way

docs/config.md
dominikg: users cannot fix these packages themselves, so we should give them an option to avoid this. 

Maybe we can change it to a list so they would still see it for newly added dependencies?

The option itself is experimental so that we can adapt on feedback

docs/faq.md
benmccann: can we change the title from `deprecated "svelte" field` to something more neutral like `prefer "svelte" export condition over "svelte" field`?

> kept the svelte field as neutral instead of remove. We should not encourage them to remove it for now as that could break older tooling that still relies on it

I agree with what you said in that comment and so the word "deprecated" seems too strong to me

packages/e2e-tests/resolve-exports-svelte/vite.config.js
benmccann: it confused me a bit when I first saw it because it made me think the default might be `true`, which is why I suggested potentially removing it

docs/config.md
benmccann: it's not very noisy since it only happens once per build. and keeping the pressure on package authors seems desirable

docs/faq.md
bluwy: If `exports` is define, Vite should ignore `mainFields` (except `browser`) completely now. Vite also doesn't have a restriction on what the extensions for `mainFields` could be, so exporting a Svelte component is fine too.

benmccann: @dominikg any thoughts about the heading title here? I think it might be the only outstanding item for us to resolve

dominikg: > @dominikg any thoughts about the heading title here? I think it might be the only outstanding item for us to resolve

i have changed the wording to use conflicting resolve results instead of deprecation.

docs/faq.md
dominikg: @dummdidumm @Rich-Harris  

ideally this faq entry contains an example of a svelte exports map that works for most/all library authors. Once we merge this PR and they start updating, many are going to have to do a breaking release. If we change the recommended setup again later that would be kind of  a d*ck move and put extra pressure on the whole ecosystem by introducing multiple major updates.

Any thoughts on how glob exports should be utilized if at all?

docs/faq.md
```diff
@@ -208,3 +208,35 @@ module.exports = {
...
 }
 ```
+
+<!-- the following header generates an anchor that is used in logging, do not modify!-->
+
+### conflicts in svelte resolve
+
+| If you see a warning logged for this when using a svelte library, please tell the library maintainers.
+
+In the past, Svelte recommended using the custom "svelte" field in package.json to allow libraries to point at .svelte source files.
```
benmccann, line 218:
```suggestion
In the past, Svelte recommended using the custom `svelte` field in `package.json` to allow libraries to point at `.svelte` source files.
```

docs/faq.md
```diff
@@ -208,3 +208,35 @@ module.exports = {
...
++   }
+  }
+```
+
+> **Support for the svelte field is going to be deprecated and later removed in a future versions of vite-plugin-svelte.**
+>
+> Library authors are highly encouraged to update their packages to add the new exports condition as outlined above. Check out
+> [svelte-package](https://kit.svelte.dev/docs/packaging) which already supports it.
+>
+> For backwards compatibility, you can keep the svelte field in addition to the exports condition. But make sure that both always resolve to the same files.
```
benmccann, line 242:
```suggestion
> For backwards compatibility, you can keep the `svelte` field in addition to the `exports` condition. But make sure that both always resolve to the same files.
```

docs/faq.md
```diff
@@ -208,3 +208,35 @@ module.exports = {
...
+
+<!-- the following header generates an anchor that is used in logging, do not modify!-->
+
+### conflicts in svelte resolve
+
+| If you see a warning logged for this when using a svelte library, please tell the library maintainers.
+
+In the past, Svelte recommended using the custom "svelte" field in package.json to allow libraries to point at .svelte source files.
+This field requires a custom implementation to resolve, so you have to use a bundler plugin and this plugin needs to implement resolving.
+Since then, node has added support for [conditional exports](https://nodejs.org/api/packages.html#conditional-exports), which have more generic support in bundlers and node itself. So to increase the compatibility with the wider ecosystem and reduce the implementation needs for current and future bundler plugins, it is recommended that packages use the "svelte" exports condition.
```
benmccann, line 220:
```suggestion
Since then, Node has added support for [conditional exports](https://nodejs.org/api/packages.html#conditional-exports), which have more generic support in bundlers and Node itself. So to increase the compatibility with the wider ecosystem and reduce the implementation needs for current and future bundler plugins, it is recommended that packages use the `svelte` exports condition.
```

docs/faq.md
benmccann: I've never seen anyone use glob exports in the wild before. It's probably not a good practice. E.g. if you have a main component that uses a few sub-components, you probably don't want to export the sub-components. I think I would change the example to not use glob exports

In terms of `index.js` vs deep package path, I don't know if we're wanting to make a universal declaration on that yet. You could potentially show both examples and then link to https://github.com/sveltejs/vite-plugin-svelte/blob/main/docs/faq.md#what-is-going-on-with-vite-and-pre-bundling-dependencies for a deeper discussion of what method to choose

dominikg: Hadn't thought about internal subcomponents. But i do think we have to provide guidance on how to allow deep imports and/or index imports. 

svelte-package stopped generating deep exports into package.json so it would be the library authors task to remember to add new components to that. But maybe thats preferable because with glob exports everything matched by the glob is public and that makes it easier to accidentally do a breaking change without realizing.

dominikg: removed the glob example and linked to the other faq entry.

docs/faq.md
```diff
@@ -208,3 +208,34 @@ module.exports = {
 +    }
 }
 ```
+
+<!-- the following header generates an anchor that is used in logging, do not modify!-->
+
+### conflicts in svelte resolve
+
+| If you see a warning logged for this when using a svelte library, please tell the library maintainers.
```
benmccann, line 216:
```suggestion
| If you see a warning logged for this when using a Svelte library, please tell the library maintainers.
```

benmccann (approved):

PR #513, opened 2022-11-29
feat: support additional queries

### LINKED ISSUE #341: Support the `direct` query param
### Describe the problem

In Astro we prefer to add `<link>` tags per component for styles in development mode. This works because Vite will detect a request that has been added via link tag through the Accept header and then injects the `direct` query param.

For Astro and Vue styles this works because those Vite plugins do not modify the id. Since Svelte creates an id the `direct` param gets lost.

This means that we cannot use `<link>` in dev for Svelte, but rather must use a module script which results in a FOUC (flash of unstyled content).

When generating the `cssId`, preserve `direct` query param, if it's included in the importee. https://github.com/sveltejs/vite-plugin-svelte/blob/a6a8e00dcd2fa05f01e7037a8062ad3afe1f7615/packages/vite-plugin-svelte/src/utils/id.ts#L52

### Alternatives considered

No alternatives that I am aware of, other than Astro continuing to use `<script>` and live with the FOUC.

----

Note that I am willing to submit a PR for this.

### Importance

would make my life easier

dominikg:
add support for ?raw and ?direct import queries.

fixes #341

> **HERE BE DRAGONS**
>
> The features described here are not meant to be used in regular libraries or end-user applications.
> They can be useful for frameworks, documentation sites or in special situations, but you are responsible for applying them correctly.
>
> **Proceed with caution!**

### raw

```js
//get .svelte file content as a string
import content from 'File.svelte?raw';

//get output of svelte.preprocess as {code, map, dependencies}
import preprocessed from 'File.svelte?raw&svelte&type=preprocessed';

//get output of svelte.compile js as {code, map, dependencies}
import script from 'File.svelte?raw&svelte&type=script';

//get output of svelte.compile css as {code, map }
import style from 'File.svelte?raw&svelte&type=style';

//get output of svelte.compile as {source, compiled:{js,css,preprocessed,dependencies,ast}}
import all from 'File.svelte?raw&svelte&type=all';
```

### direct

```html
<!-- load and execute component script -->
<script type="application/javascript" src="File.svelte?direct&svelte&type=script&lang.js" />
<!-- embed component style as css -->
<link rel="stylesheet" type="text/css" href="File.svelte?direct&svelte&type=style&lang.css" />
```


bluwy (approved):
LGTM. For `?raw` with `&type` etc, I think we should rename `?raw` to something else as Vite defines it as something that returns a string. Right now it returns an object with `code` and `map` where they can't be treeshaken if you use code only.

This could be done as another PR though as it's experimental.

bluwy (approved):
Mixed exports 🚀

PR #522, opened 2022-12-06
feat!: remove commonjs build

### LINKED ISSUE #487: remove cjs build of vite-plugin-svelte
### Describe the problem

currently v-p-s is built as a dual-module supporting both esm and cjs. Svelte projects are typically `type: module` so esm is fine for regular use. The cjs output was only added because initially our test-suite used jest and that didn't support esm.

The dual out requires utility libraries like vitefu to also offer cjs out and it bloats our package.

remove the output and go esm only

### Alternatives considered

keep it, at the expense of everyone

### Importance

would make my life easier

dominikg:
fixes #487 

see [faq](https://github.com/sveltejs/vite-plugin-svelte/blob/main/docs/faq.md#how-can-i-use-vite-plugin-svelte-from-commonjs) if you need commonjs


bluwy (approved):

bluwy:
side topic, i'm not sure if `breaking:` is a valid scope. i usually see it as `feat!:` or `chore!:` 🤔

PR #546, opened 2022-12-11
fix: kit compatibility

dominikg:
1) detect kit from presence of it's vite plugin for inspector.

kit-node test still fails due to config file changes not being tracked, have to investigate


packages/vite-plugin-svelte/src/ui/inspector/plugin.ts
```diff
@@ -35,26 +35,28 @@ export function svelteInspector(): Plugin {
...
+				// but that should be rare and inspector doesn't execute twice
+				inspectorOptions.appendTo = `/generated/root.svelte`;
+			}
+			appendTo = inspectorOptions.appendTo;
 		},
 
 		async resolveId(importee: string, importer, options) {
-			if (options?.ssr || disabled) {
+			// @ts-expect-error scan is experimental
+			if (options?.ssr || options?.scan || disabled) {
```
bluwy, line 59: Is this scan check required? Or perhaps for optimisation only? Would be nice to avoid it since it's internal 🤔

dominikg: removed :+1:

bluwy (approved):

PR #554, opened 2022-12-13
Make Inspector Toggle draggable

LorisSigrist:
https://user-images.githubusercontent.com/43482866/207358315-00568bca-6e69-478a-8e6d-aba7250033a0.mov

## The Issue 
Currently developers have four choices for where to place the Toggle-Button for the svelte inspector; The four corners.
Since the toggle-position is set for the entire site, having an element in a corner on any page makes that corner undesirable for the inspector button. Its pretty easy to run into a situation where no corner is a good option for the toggle.

## The Fix
Make the Toggle Button draggable to anywhere on the screen.
 
## This PR
- Makes the Toggle Draggable
- Makes sure the Toggle stays on screen, even if the window resizes
- Persists the position of the toggle in localstorage
- Removes the `toggleButtonPos` field from the config and docs as it is no longer needed

I tested this in Chrome, Firefox and Safari on a Mac, aswell as on an android phone. Seems to work.

### Disclaimer
There is a bug where if you toggle the inspector on, you are able to select the toggle itself as an element, which opens the root layout. Once your cursor leaves the toggle you wont be able to select it again. This bug is **not** introduced by this PR, it existed previously, but it is a lot easier to run into it. Not sure why that happens though.


dominikg:
Hi, this is an interesting proposal.

Improving the positioning of the button is a good idea, however i don't think dragging is needed and in the current implementation it adds to much to the running application.

Neither listeners nor localstorage should be needed for this, just some custom css. Either by targeting it's css class in a dev-time only stylesheet or we could look into adding an option to customize the styles from within the inspector config.

LorisSigrist:
Ok, that makes sense. Thank you!
I closed the pull requests as to not clutter up the repo

PR #555, opened 2022-12-15
Fixes: Inspector doesn't work with firefox on Linux because of 'meta-shift' as default for all platforms but Windows

delight:
The 'meta-shift' shortcut won't work on firefox when used on linux. Also in KDE the shortcut is preset to draw mouse trails on the screen. Switching to control-shift on linux but is a little tricke to work with, because of the way it handles the click to activate the browser window again.

Having tried all sort of options/combinations it seems the best working for all browsers
is the combination of "control-shift" and holdMode: true

_**I'm stil wondering**_  if instead of checking
['linux', 'win32'].includes(process.platform) ? 'control-shift' : 'meta-shift',

it would make more sense to check just for mac by doing a 

process.platform === 'darvin' ? 'meta-shift' : 'control-shift',

for sure it is cumbersome having to change the options in svelte.config.js manually every time when switching the platform (mac <-> linux)

Discussed some on:
https://discord.com/channels/457912077277855764/1052953876724207676

Note: The preset of meta-shift can be deactivated in systemsettings of KDE but firefox still won't work with it.

### DIFF
--- a/.changeset/gorgeous-peas-mix.md
+++ b/.changeset/gorgeous-peas-mix.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/vite-plugin-svelte': minor
+---
+
+makes 'meta-shift' the exceptin for the mac - fixes the problem on Linux / firefox etc.
--- a/docs/config.md
+++ b/docs/config.md
@@ -286,7 +286,7 @@ export default {
   interface InspectorOptions {
     /**
      * define a key combo to toggle inspector,
-     * @default 'control-shift' on windows, 'meta-shift' on other os
+     * @default 'meta-shift' on mac, 'control-shift' on other os
      *
      * any number of modifiers `control` `shift` `alt` `meta` followed by zero or one regular key, separated by -
      * examples: control-shift, control-o, control-alt-s  meta-x control-meta
--- a/packages/vite-plugin-svelte/src/ui/inspector/plugin.ts
+++ b/packages/vite-plugin-svelte/src/ui/inspector/plugin.ts
@@ -7,10 +7,10 @@ import fs from 'fs';
 import { idToFile } from './utils';
 
 const defaultInspectorOptions: InspectorOptions = {
-	toggleKeyCombo: process.platform === 'win32' ? 'control-shift' : 'meta-shift',
+	toggleKeyCombo: ['linux', 'win32'].includes(process.platform) ? 'control-shift' : 'meta-shift',
 	navKeys: { parent: 'ArrowUp', child: 'ArrowDown', next: 'ArrowRight', prev: 'ArrowLeft' },
 	openKey: 'Enter',
-	holdMode: false,
+	holdMode: process.platform === 'linux',
 	showToggleButton: 'active',
 	toggleButtonPos: 'top-right',
 	customStyles: true
--- a/packages/vite-plugin-svelte/src/ui/inspector/plugin.ts
+++ b/packages/vite-plugin-svelte/src/ui/inspector/plugin.ts
@@ -7,10 +7,10 @@ import fs from 'fs';
 import { idToFile } from './utils';
 
 const defaultInspectorOptions: InspectorOptions = {
-	toggleKeyCombo: ['linux', 'win32'].includes(process.platform) ? 'control-shift' : 'meta-shift',
+	toggleKeyCombo: process.platform === 'darwin' ? 'meta-shift' : 'control-shift',
 	navKeys: { parent: 'ArrowUp', child: 'ArrowDown', next: 'ArrowRight', prev: 'ArrowLeft' },
 	openKey: 'Enter',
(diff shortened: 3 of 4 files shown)


dominikg:
unfortunately there is no perfect key-combo for all platforms and browsers. meta-shift works on linux chromium based browsers.

I'd prefer we keep the current defaults. You could PR an addendum to the docs that gets into more detail about fine-tuning the combo for your needs.

In multi-environment setups, you could use `process.env.SOME_ENV='my-combo'` and `keyCombo: process.env.SOME_ENV ?? 'fallback-combo'`  to allow yourself or team members to override the default to something that works for them.

In the same direction I don't think holdMode should be changed to on by default on a single platform. If at all it should default to on for all platforms if thats what most users prefer.

delight:
> unfortunately there is no perfect key-combo for all platforms and browsers. meta-shift works on linux chromium based browsers.
> 
> I'd prefer we keep the current defaults. You could PR an addendum to the docs that gets into more detail about fine-tuning the combo for your needs.
> 
> In multi-environment setups, you could use `process.env.SOME_ENV='my-combo'` and `keyCombo: process.env.SOME_ENV ?? 'fallback-combo'` to allow yourself or team members to override the default to something that works for them.
> 
> In the same direction I don't think holdMode should be changed to on by default on a single platform. If at all it should default to on for all platforms if thats what most users prefer.

I agree on the holdMode. 

About the shortcut though, right now it is a workaround for mac based computers, crippling the functionality on other OS. As mentioned earlier meta-shift is a pre-registered combination on some DEs. 'control+shift' works for all browsers - with or without holdMode: true. holdMode : true just gave a better user experience.

I'll modify the patch accordingly to keep 'meta-shift'  as the exception just for the mac. That way it will work on more platforms and browsers including firefox + it won't be blocked by the DEs global shortcuts (e.g. KDE).

holdMode: true is a big :+1:  from me. But for sure it depends on what the majority of the user wants.

delight:
Now this is probably way better and cleaner: 

Now you could simply:

```ts
	vitePlugin: {
		experimental: {
			inspector: {
				holdMode: true,
			}
		}
	}
```

where the holdmode is obiously a choice and have it working on all platforms. Surely it remains tricky and you'd need to find ppl to try it on more exotic platforms and see if there is any bug remaining.

delight:
I'd agree in general but argue that it doesn't work at all for many linux users. 1 or 2 days ago I was looking for the this repo and asked in the community chat. There were 2 or 3 more ppl using kde and it didn't work at all for them They were wondering about it. Also firefox is not working either. 

I'd also argue as it is in experimental and it is "just" the linux users that get affected the harm might be smaller << but I guess that is a small consolation.

The environment env sounds like a feasible option. I'd still argue that ctrl+shift is more natural to most users inlcuding linux but ok.

summing up:
#### contra meta-shift on  linux
- firefox not working
-  kde has a global shortcut, blocking it

#### pro(meta-shift on linux
- it is the default since setting the workaround for mac
- muscle memory of Linux users.

Me switching OS and computers often somehow don't worry about it as much and I keep adapting quickly, but that is just me. 

###### Ultimately it all depends on documentation + I guess maybe it would be helpful to write a index.d.ts file for the plugins config options to help with intellisense and provide documentation inline.

delight:
- meta-shift doesn't play along well with edge on linux

benmccann:
`meta-shift` does nothing for me in Chrome on Linux. I could only get `alt-shift` and `control-shift` to work.

I don't know what `meta` even is. MDN seems to indicate [the key would say "meta"](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/getModifierState), but I don't have such a key on my keyboard. I have a Windows key on the external keyboard I use, which doesn't seem to do anything in relation to inspector. I tried both `meta-shift` and `os-shift` as [I believe the latter is what would be triggered](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/getModifierState). Neither did anything. I needed some code changes to test `os-shift` as inspector does not currently support it and it's possible I didn't implement it properly.

I also had a terrible time figuring out that I needed to type `control` rather than `ctrl`. It would be nice to accept both variants.

dominikg:
While we could add alias support for some keys, i think it's best to stick to the key event names. `control` is also heavily documented here https://github.com/sveltejs/vite-plugin-svelte/blob/main/docs/config.md#inspector

"meta" is triggered for me on manjaro-linux chromium by pressing the windows logo key on my keyboard. Not sure why it doesn't for you, but if you want to add support for `os`, i think it would need to be inserted in this switch and supported by getModifierState https://github.com/sveltejs/vite-plugin-svelte/blob/main/packages/vite-plugin-svelte/src/ui/inspector/Inspector.svelte#L134  (or have name === key, in which case it would already work)

dominikg:
closing in favor of #631  which also adds environment configuration

Issue #578, opened 2023-01-22
vitePreprocess runs Sass, but PostCSS too


bfanger:
### Describe the bug

When using `<style lang="scss">` vitePreprocess also applies the PostCSS transformations.

Take the following example:
```html
<button class="btn">Hello</button>

<style lang="scss">
  .btn {
    background: gray;
    border: 2px outset;

    &:focus-visible {
      outline: 2px solid orange;
    }
  }
</style>
```

When using postcss-preset-env this results in:  
 `[vite-plugin-svelte] /src/App.svelte:8:0 Unused CSS selector ".btn.focus-visible"`

It looks like the transformations are:

| Step |  Processor | Effect |
|--|--|--|
|1| Sass | creates `.btn:focus-visible` selector |
|2| PostCSS | adds `.btn.focus-visible` selector
|3| Svelte Compiler | removes `.btn.focus-visible` selector
|4| PostCSS | adds the `.btn.focus-visible selector` again

I think the default should be to remove the 2nd step, don't run PostCSS before passing the css to the Svelte Compiler.

### Reproduction URL

https://stackblitz.com/edit/vitejs-vite-pkrz2a?file=src/App.svelte

### Reproduction

The error is visible in the console when the App.svelte is compiled.

1. `npm run dev`

### Logs

```shell
[vite-plugin-svelte] /src/App.svelte:8:0 Unused CSS selector ".btn.focus-visible"
```

### System Info

```shell
Binaries:
    Node: 16.14.2 - /usr/local/bin/node
  npmPackages:
    @sveltejs/vite-plugin-svelte: ^2.0.2 => 2.0.2 
    svelte: ^3.55.1 => 3.55.1 
    vite: 4.0.4
```

dominikg:
is the behavior different from using svelte-preprocess?

vitePreprocess applies vites css pipeline before compile, I'm not sure you can selectively disable parts of it

bluwy:
Hmm I didn't take into account that Vite would also apply postcss for normal css files. I'd say this is expected behaviour as Vite does this too, and could be worked around by suppressing the warning with onwarn.

Maybe we should also let Vite preprocess regardless if `lang` exists in this case 🤔

dominikg:
While the double processing is unfortunate and we should look into possible ways to avoid it, i wonder why it creates a class for focus-visible in the first place. It has really good browser support nowadays, so unless you target obscure or dead browsers, it should not even turn that into an extra class.

[...]

PR #580, opened 2023-01-24
fix: Don't use PostCSS config file when preprocessing other languages

bfanger:
When the **lang** attribute of the **style** block isn't **postcss**, that means another preprocessor is wanted for the preprocessing, like sass for example `<style lang="scss">`

But when a postcss config is found the `preprocessCSS()` from vite will load and apply PostCSS as well, which leads to bugs/warnings  #578

This PR adds an inline postcss config to prevent loading the postcss config.
It also contains one dummy postcss plugin, this triggers the code inside vite to fix the sourcemapping, preventing

`Sourcemap for "path/to/file.svelte" points to missing source files` warnings

### DIFF
--- a/packages/vite-plugin-svelte/src/preprocess.ts
+++ b/packages/vite-plugin-svelte/src/preprocess.ts
@@ -67,6 +67,17 @@ function viteStyle(config: InlineConfig | ResolvedConfig = {}): {
 					process.env.NODE_ENV === 'production' ? 'build' : 'serve'
 				);
 			}
+			if (lang !== 'postcss') {
+				resolvedConfig = {
+					...resolvedConfig,
+					css: {
+						...(resolvedConfig.css ?? {}),
+						postcss: {
+							plugins: [{ postcssPlugin: 'vitePreprocess' }]
+						}
+					}
+				};
+			}
 			transform = getCssTransformFn(resolvedConfig);
 		}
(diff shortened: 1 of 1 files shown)


dominikg:
Unfortunately this change would be a breaking change and it does not work correctly in all cases.

Depending on the postcss plugins present in the user config, the user also might need/want to apply postcss transforms for lang=scss, eg if they use tailwind. Also processing of `url()` values is needed if the css isn't externalized into a separate file and as a result never processed by vites postcss plugin again.

The way the transform function is cached right now would mean that the first encountered file would determine if postcss is used or not.  It would have to be changed to cache a transform function for each lang attribute (or group several that can share one).

The only way this could be introduced to vite-plugin-svelte 2.x is with a new option for vitePreprocess, eg `skipPostcss: boolean` that defaults to false. 

But all this effort seems a bit much for me to avoid a warning that likely goes away too if you configure browserslist correctly (see my comment in #578)

bfanger:
This PR doesn't disable PostCSS entirely. PostCSS **still** runs with the plugins specified in the config, so all PostCSS transforms like tailwind should still work as expected.

The transform is indeed cached, now you can't use `lang="postcss"` inside a project that also uses other processors, i'll fix that tomorrow.

dominikg:
see #578

PR #602, opened 2023-03-10
chore: update svelte to 3.56.0 and update test snapshots 

dominikg:
this fixes vite-ecosystem-ci failing because that always runs with `"svelte":"latest"`


dominikg:
change in svelte 3.56.0 that required updating the snapshot https://github.com/sveltejs/svelte/issues/8134

bluwy (approved):

PR #614, opened 2023-04-07
fix: log stats in debug mode

bluwy:
Log client/server compile stats in debug mode instead of info. This has been bugging me for a while since it shows up in debug and CI logs by default, making debugging hard 😅 

IIUC it's shown so that users understand what's taking so long to compile/prebundle. I checked Vite 4.3 with prebundling, and it only had a 1.5s improvement (10.5s in Vite 4.3). The bottleneck seems to still be in the Svelte compiler according to the cpuprofile.


dominikg:
hmm, you can disable the compile stats in dev/build or completely. The purpose of the info log was mostly to make this information accessible to regular users as debug is very noisy.

Even with vite's recent enhancements to performance tracking, it would be hard to get information aggregated by svelte dependency out of it. 

But I did not receive any feedback from the logged stats, so if you think this is detrimental to do by default, maybe instead of moving it do debug level, we should just switch the default value of the config flag to true (or rather rename it to enableCompileStats)?

https://github.com/sveltejs/vite-plugin-svelte/blob/main/docs/config.md#disablecompilestats

bluwy:
Maybe we can expose the stats under a specific namespace like `vite:vite-plugin-svelte:stats`? Then we can have someone run with `DEBUG="vite:vite-plugin-svelte:stats"` to achieve the same result.

I'm not sure about showing the stats by default since it isn't always useful for user's projects, as it's only helpful for large Svelte deps, or large Svelte SSR files. Even then with the info presented, the user might not be looking into fixing it just yet. If they have to enable `DEBUG=` manually, it likely means they're willing to take time to optimize things.

bluwy:
Updated to use a namespace:
```
Forced re-optimization of dependencies

  VITE v4.2.1  ready in 455 ms

  ➜  Local:   http://127.0.0.1:5173/
  ➜  press h to show help
  vite:vite-plugin-svelte:stats ssr compile done.
  vite:vite-plugin-svelte:stats package                 files     time     avg
  vite:vite-plugin-svelte:stats playground-kit-demo-app     5   54.9ms  11.0ms +0ms
```

```
  VITE v4.2.1  ready in 321 ms

  ➜  Local:   http://127.0.0.1:5173/
  ➜  press h to show help
  vite:vite-plugin-svelte:stats prebundle libraries in progress ... +0ms
  vite:vite-plugin-svelte:stats prebundle libraries done.
  vite:vite-plugin-svelte:stats package                         files    time     avg
  vite:vite-plugin-svelte:stats carbon-icons-svelte              2002   7.22s   3.6ms
  vite:vite-plugin-svelte:stats carbon-components-svelte          206   1.86s   9.0ms +12s
```

packages/vite-plugin-svelte/src/utils/options.ts
```diff
@@ -207,10 +207,7 @@ export function resolveOptions(
 	enforceOptionsForProduction(merged);
 	// mergeConfigs would mangle functions on the stats class, so do this afterwards
 	const isLogLevelInfo = [undefined, 'info'].includes(viteConfig.logLevel);
-	const disableCompileStats = merged.experimental?.disableCompileStats;
-	const statsEnabled =
-		disableCompileStats !== true && disableCompileStats !== (merged.isBuild ? 'build' : 'dev');
-	if (statsEnabled && isLogLevelInfo) {
+	if (isLogLevelInfo) {
```
dominikg, line 210: removing the statsEnabled flag here would cause the stat collection to always happen unless user sets loglevel to warning or above, which would be out of sync with the change to loglevel debug for stat output.

I'm still not sure why moving the logging to a debug namespace is better than flipping the default value of the stats enabled option.

bluwy: Updated to check based on `log.debug.enabled` (https://github.com/sveltejs/vite-plugin-svelte/pull/614/commits/59a1034bfe681c8da10b31185d63fb4f65bfe078).

I guess for me the lesser the options the better. And logging stats is kind of a debug info that should be tucked with other debug logs.

.changeset/strong-coins-shout.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sveltejs/vite-plugin-svelte': patch
+---
+
+Log stats in debug mode
```
dominikg, line 5:
```suggestion
Log stats in debug mode  and remove `experimental.disableCompileStats` option
```

bluwy:
```suggestion
Log stats in debug mode and remove `experimental.disableCompileStats` option. Use `DEBUG="vite:vite-plugin-svelte:stats"` when starting the dev server or build to log the compile stats.
```

dominikg (approved):

PR #615, opened 2023-04-10
refactor: use generic namespace for logging instead of hardcoded stats

dominikg:
more generic version for namespaced log/debug, including an improved check if the stats namespace has been enabled by the user


bluwy (approved):
neato

PR #618, opened 2023-04-13
chore: update pnpm 8

dominikg:


dominikg:
looks like the file: protocol in for _test_dependencies works differently in pnpm@8 :/

bluwy (approved):

bluwy:
Ah wait I approved without reading the comments 🙈

benmccann:
Should we use `pnpm/action-setup` in the CI like SvelteKit does? I noticed this repo has some custom logic currently which could be simplified

package.json
```diff
@@ -55,11 +55,8 @@
       "prettier --write"
     ]
   },
-  "packageManager": "pnpm@7.30.0",
+  "packageManager": "pnpm@8.2.0",
   "engines": {
-    "pnpm": "^7.13.0",
-    "yarn": "forbidden, use pnpm",
-    "npm": "forbidden, use pnpm",
```
benmccann, line 62: you can leave this in place if you'd like. you just need to switch `^` to `>=` which I believe is the thing that's currently stopping me from using this repo

.github/workflows/ci.yml
```diff
@@ -87,11 +87,15 @@ jobs:
         with:
           node-version: ${{ matrix.node }}
       - name: install pnpm
+        if: matrix.node != 14
         shell: bash
         run: |
           PNPM_VER=$(jq -r '.packageManager | if .[0:5] == "pnpm@" then .[5:] else "packageManager in package.json does not start with pnpm@\n" | halt_error(1)  end' package.json)
           echo installing pnpm version $PNPM_VER
           npm i -g pnpm@$PNPM_VER
+      - name: install legacy pnpm for node14
```
benmccann, line 96: you could just leave `packageManager` as `7.32.0` in `package.json` and then you wouldn't need to do any of this. that's much simpler. people can still have pnpm 8 in their local installs, which is what's really important

dominikg:
> Should we use `pnpm/action-setup` in the CI like SvelteKit does? I noticed this repo has some custom logic currently which could be simplified

that action uses curlbashing and it also adds an additional point where pnpm version is set. this setup here uses packageManager as canonical and doesnt require a separate action. ymmv but i very much prefer the current approach

.github/workflows/ci.yml
dominikg: the goal is that everyone uses the same major version of the same pm. differences between pnpm 7 and 8 make it undesirable to allow contributing with both, as evidenced by the current problem of file: resolution in this PRs test run

package.json
dominikg: you can enable corepack locally to ease working with different package manager versions, or switch manually. until the issue with 8 is resolved, 7 must be used.

benmccann:
Just FYI, the action also uses `packageManager` and doesn't require setting the version used separately from that

package.json
dominikg: added back engines.pnpm and rewriting it for node14 tests. left the other two removed as neither npm nor yarn check it anymore i think.

bluwy (approved):

PR #621, opened 2023-04-14
fix: Sourcemap for "/path/to/Component.svelte" points to missing source files

bfanger:
When using Vite's [css.devSourcemap: true](https://vitejs.dev/config/shared-options.html#css-devsourcemap) and using `<style lang="scss">` processed with `vitePreprocess` will result in broken source maps for the css.

Example vite log:
```
Sourcemap for "/Volumes/Sites/svelte-project/src/routes/+page.svelte" points to missing source files
```

This is caused by the the fake filename that is passed to the `transform` function (which is needed so vite will select the appropriate transformation).

This PR repairs the SourceMap to use the original filename.

### DIFF
--- a/packages/vite-plugin-svelte/src/preprocess.ts
+++ b/packages/vite-plugin-svelte/src/preprocess.ts
@@ -1,3 +1,4 @@
+import path from 'path';
 import { preprocessCSS, resolveConfig, transformWithEsbuild } from 'vite';
 import type { ESBuildOptions, InlineConfig, ResolvedConfig } from 'vite';
 // eslint-disable-next-line node/no-missing-import
@@ -74,9 +75,19 @@ function viteStyle(config: InlineConfig | ResolvedConfig = {}): {
 		}
 		const moduleId = `${filename}.${lang}`;
 		const { code, map } = await transform(content, moduleId);
-
 		mapSourcesToRelative(map, moduleId);
-
+		const baseModuleId = path.basename(moduleId);
+		const baseFilename = path.basename(filename);
+		if (map?.file === moduleId) {
+			map.file = filename;
+		} else if (map?.file === baseModuleId) {
+			map.file = baseFilename;
+		}
+		if (map?.sources?.map) {
+			map.sources = map.sources.map((source: string) =>
+				source === baseModuleId ? baseFilename : source
+			);
+		}
 		return {
 			code,
 			map: map ?? undefined
--- a/.changeset/real-jokes-help.md
+++ b/.changeset/real-jokes-help.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/vite-plugin-svelte': patch
+---
+
+fixed Sourcemap for "Component.svelte" points to missing source files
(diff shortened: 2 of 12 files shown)


packages/vite-plugin-svelte/src/preprocess.ts
```diff
@@ -76,7 +77,14 @@ function viteStyle(config: InlineConfig | ResolvedConfig = {}): {
 		const { code, map } = await transform(content, moduleId);
 
 		mapSourcesToRelative(map, moduleId);
-
+		if (map?.file === moduleId) {
+			map.file = filename;
+		}
+		if (map?.sources?.map) {
```
dominikg, line 83: wouldn't this have to be called before mapSourcesToRelative (and then use filename instead of moduleId as second arg for that)

packages/vite-plugin-svelte/src/preprocess.ts
```diff
@@ -76,7 +77,14 @@ function viteStyle(config: InlineConfig | ResolvedConfig = {}): {
...
+			map.sources = map.sources.map((source: string) =>
+				source === path.basename(moduleId) ? path.basename(filename) : source
```
dominikg, line 85: both basename calls can be moved out of the map call, though i'm not sure if this is needed anymore if the code is moved above the mapSourcesToRelative call

dominikg:
thank you for posting this. Please add a changeset with `pnpm changeset` and it would also be great if there was at least a reproduction repo or even better a testcase that shows this is working as intended afterwards.

packages/vite-plugin-svelte/src/preprocess.ts
bfanger: In my use-case the entry in sources was the absolute path that matches the moduleId.

bfanger:
Interesting, the sourcemap from the unittest i'm writing does use relative paths and a sourceRoot before it is processed by the `mapSourcesToRelative()`:
```
{
  version: 3,
  sourceRoot: '/Volumes/Sites/vite-plugin-svelte/packages/e2e-tests/css-dev-sourcemap/src',
  sources: [ 'App.svelte.scss' ],
  names: [],
  mappings: 'AAEE;EACC',
  file: 'App.svelte.scss'
}
```

bfanger:
I've added an e2e test, so it keeps working  as intended #305 😉.  
Feel free to contact me when new SourceMap issues arise. 

**Tip:** If you paste the contents of the snapshot from  
`packages/e2e-tests/css-dev-sourcemap/__tests__/__snapshots__/css-dev-sourcemap.spec.ts.snap`
Into the  [Source Map Visualization tool](https://evanw.github.io/source-map-visualization) you'll see it now working as intended:
source-map-visualization

dominikg:
Thanks for adding the testcase. I'm not sure using a sourcemap visualizer to be able to understand the snapshot is a good idea as it puts extra work on the maintainers every time the snapshot changes to find out why exactly it suddenly failed.

A unit test calling vitePreprocess().style directly could be better.

The filenames being relative already could be vite having improved in the meantime or because this is css and it's only a simple input. If there was additional config, eg a banner or scss imports leading to more entries in map.sources, the situation could be different.

For this PR I think it is sufficient to make sure that the extra lang suffix is stripped from the file refs in map

dominikg:
refactored it here https://github.com/sveltejs/vite-plugin-svelte/pull/625 . @bfanger  please take a look. It should cover your case and then some (eg turning map.file into relative as well)

bfanger:
Closing in favor of the 625

PR #622, opened 2023-04-14
Skip Vite resolve workaround on Vite 4.1+ or Svelte 4+

benmccann:
[according to @dominikg](https://github.com/sveltejs/svelte/pull/8497#issuecomment-1506354874):

> latest vite does not need the custom resolve for it in v-p-s anymore, but early vite 4 versions did

I think we can safely remove this. People who are on old versions of Vite 4 may get larger apps if they upgrade `vite-plugin-svelte` without upgrading Vite, but that's relatively safe as nothing will break.

There's some nice benefits to removing this including simplifying the `vite-plugin-svelte` code and eventually paving the way for us to stop exporting `svelte/ssr` in Svelte core

### DIFF
--- a/.changeset/chilled-tigers-flash.md
+++ b/.changeset/chilled-tigers-flash.md
@@ -2,4 +2,4 @@
 '@sveltejs/vite-plugin-svelte': patch
 ---
 
-Skip Vite resolve workaround on Vite 4.1+
+Skip Vite resolve workaround on Vite 4.1+ or Svelte 4+
--- a/.changeset/chilled-tigers-flash.md
+++ b/.changeset/chilled-tigers-flash.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/vite-plugin-svelte': patch
+---
+
+Skip Vite resolve workaround on Vite 4.1+
--- a/packages/vite-plugin-svelte/src/index.ts
+++ b/packages/vite-plugin-svelte/src/index.ts
@@ -1,5 +1,13 @@
 import fs from 'fs';
-import { HmrContext, ModuleNode, Plugin, ResolvedConfig, UserConfig, version as viteVersion } from 'vite';
+import { VERSION as svelteVersion } from 'svelte/compiler';
+import {
+	HmrContext,
+	ModuleNode,
+	Plugin,
+	ResolvedConfig,
+	UserConfig,
+	version as viteVersion
+} from 'vite';
 // eslint-disable-next-line node/no-missing-import
 import { isDepExcluded } from 'vitefu';
 import { handleHotUpdate } from './handle-hot-update';
@@ -135,8 +143,13 @@ export function svelte(inlineOptions?: Partial<Options>): Plugin[] {
 					}
 				}
 
-				// TODO: remove this after bumping peerDep on Vite to at least 4.1
-				if (viteVersion.startsWith('4.0') && ssr && importee === 'svelte') {
+				// TODO: remove this after bumping peerDep on Vite to 4.1+ or Svelte to 4.0+
+				if (
+					viteVersion.startsWith('4.0') &&
+					svelteVersion.startsWith('3') &&
+					ssr &&
+					importee === 'svelte'
+				) {
 					if (!resolvedSvelteSSR) {
 						resolvedSvelteSSR = this.resolve('svelte/ssr', undefined, { skipSelf: true }).then(
 							(svelteSSR) => {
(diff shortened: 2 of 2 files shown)


dominikg:
This requires a changeset and is a breaking change as we would have to bump the peer dependency on vite.

dominikg:
Addendum: If you remove the svelte/ssr export that contains the onMount as noop. The feature would stop working altogether.

benmccann:
I can't currently run `pnpm changeset`, but will generate a changeset after https://github.com/sveltejs/vite-plugin-svelte/pull/623 is merged.

I don't think it's a breaking change because nothing breaks if this is removed. The user just ends up with a slightly larger bundle, but that doesn't seem breaking to me.

Removing `svelte/ssr` would be breaking. I think the sooner we get this change in the better so that one day in the future when we do decide we want to remove `svelte/ssr` most users will have upgraded this library already.

dominikg:
slightly larger could be significantly larger and drive an edge deployment over a limit. I am not willing to take that risk.
If you want the standard resolve to be preferred, it would have to be with a major or fallback code for older vite-4

benmccann:
> fallback code for older vite-4

Do you know what version of Vite 4 this was fixed in or have a link to the issue this was working around? I wouldn't know where to start looking

bluwy:
It's fixed in https://github.com/vitejs/vite/pull/11595, which is released in Vite 4.1

benmccann:
I've updated this to only do the resolve workaround on Vite 4.0

bluwy (approved):

packages/vite-plugin-svelte/src/index.ts
```diff
@@ -135,7 +135,9 @@ export function svelte(inlineOptions?: Partial<Options>): Plugin[] {
 					}
 				}
 
-				if (ssr && importee === 'svelte') {
+				// TODO: remove this after bumping peerDep on Vite to at least 4.1
+				const vers = version.split('.').map((s) => parseInt(s));
```
dominikg, line 139: this can fail with a beta version, a simple startsWith 4.0 is shorter and safer

packages/vite-plugin-svelte/src/index.ts
```diff
@@ -1,5 +1,5 @@
 import fs from 'fs';
-import { HmrContext, ModuleNode, Plugin, ResolvedConfig, UserConfig } from 'vite';
+import { HmrContext, ModuleNode, Plugin, ResolvedConfig, UserConfig, version } from 'vite';
```
dominikg, line 2:
```suggestion
import { HmrContext, ModuleNode, Plugin, ResolvedConfig, UserConfig, version as viteVersion } from 'vite';
```

packages/vite-plugin-svelte/src/index.ts
dominikg, line 139:
```suggestion
```

packages/vite-plugin-svelte/src/index.ts
```diff
@@ -135,7 +135,9 @@ export function svelte(inlineOptions?: Partial<Options>): Plugin[] {
 					}
 				}
 
-				if (ssr && importee === 'svelte') {
+				// TODO: remove this after bumping peerDep on Vite to at least 4.1
+				const vers = version.split('.').map((s) => parseInt(s));
+				if (vers[0] === 4 && vers[1] === 0 && ssr && importee === 'svelte') {
```
dominikg, line 140:
```suggestion
				if (viteVersion.startsWith('4.0') && ssr && importee === 'svelte') {
```

.changeset/chilled-tigers-flash.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sveltejs/vite-plugin-svelte': patch
+---
+
+Skip Vite resolve workaround on Vite 4.1+
```
benmccann, line 5:
```suggestion
Skip Vite resolve workaround on Vite 4.1+ or Svelte 4+
```

benmccann:
I updated this to check the Svelte version as well

packages/vite-plugin-svelte/src/index.ts
```diff
@@ -135,7 +143,13 @@ export function svelte(inlineOptions?: Partial<Options>): Plugin[] {
 					}
 				}
 
-				if (ssr && importee === 'svelte') {
+				// TODO: remove this after bumping peerDep on Vite to 4.1+ or Svelte to 4.0+
+				if (
+					viteVersion.startsWith('4.0') &&
+					svelteVersion.startsWith('3') &&
+					ssr &&
+					importee === 'svelte'
```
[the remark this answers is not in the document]
dominikg: or move the startsWith results to consts?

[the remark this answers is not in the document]
benmccann: I moved the `startWith` to `const`s. I think it will be more performant to do the Vite comparison first because then a single comparison will usually be enough. If we do the `ssr` comparison first it will be true at most half a time requiring additional comparisons.

dominikg (approved):

PR #624, opened 2023-04-14
Allow pnpm 7 or 8

benmccann:
This is a simpler alternative to https://github.com/sveltejs/vite-plugin-svelte/pull/618, which would unblock people immediately without needing to address any of the test compatibility issues with pnpm 8 and avoids the CI complication of using different versions of pnpm with different versions of Node

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,7 +57,7 @@
   },
   "packageManager": "pnpm@7.30.0",
   "engines": {
-    "pnpm": "^7.13.0",
+    "pnpm": ">= 7.13.0",
     "yarn": "forbidden, use pnpm",
     "npm": "forbidden, use pnpm",
     "node": "^14.18.0 || >= 16"


dominikg:
As mentioned in another PR, i don't think it is a good idea to allow working on a project with different major versions of a package manager. I've been burned by multiple migrations from npm x to npm y, and pnpm 6 to 7 wasn't a walk in the park either.

The exception for using node14 with pnpm7 is needed because pnpm8 doesn't work with node14 and we still need to run the CI on 14 until we are able to drop it in the next major (probably alongside vite 5 in fall, maybe earlier if a new svelte release mandates it)

benmccann:
Yeah, pnpm 6 and 7 was terrible. I think they learned from that and made compatibility better with 7 and 8 though. I've got other projects setup this way and haven't hit any issues.

benmccann:
Ah, I hadn't seen your response in the other thread yet. I guess the test issues in that PR are a fair example of where 7 and 8 differ

PR #559, opened 2022-12-16
feat(inspector): make holdMode the new default

delight:
According to the discussion on, let's try holdMode as the new default as long as the plugin is experimental

https://discord.com/channels/457912077277855764/1053296007246516314

PR #631, opened 2023-04-23
feat: svelte inspector updates and promote from experimental

### LINKED ISSUE #434: Inspector doesn't work with modified app.html
### Describe the bug

In my sveltekit project, I (unwisely) altered part of the ```app.html``` from
```html
<div>%sveltekit.body%</div>
```
To
```html
%sveltekit.body%
```
After spending an hour or two, I realised that this was the source of the problem, and that clearly the whole point of %sveltekit.body% being in a container ```<div>``` is to allow other scripts (such as inspector) to add new elements to <body> without them getting over-written.
Since I believe that my misunderstanding is not uncommon, it would be useful if inspector could spot the issue and error or warn as appropriate.
Perhaps in ```create_inspector_host()``` a MutationObserver could be used to spot if the svelte-inspector-host ```<div>``` is removed, and produce a console error (or maybe there is a better way?)

### Reproduction

Create new sveltekit app
Alter this line in ````app.html``` from
```html
<div>%sveltekit.body%</div>
```
to
```html
%sveltekit.body%
```
Enable inspector by adding to svelte.config.js:
```js
vitePlugin: {
	experimental: {
		inspector: {
			showToggleButton: "always",
			toggleButtonPos: "top-right",
		},
	},
},
```
Start dev server, and notice that inspector isn't working

### Logs

_No response_

### System Info

```shell
System:
    OS: Windows 10 10.0.19044
    CPU: (4) x64 Intel(R) Core(TM) i5-4670K CPU @ 3.40GHz
    Memory: 8.18 GB / 15.93 GB
  Binaries:
    Node: 16.16.0 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.13.0 - C:\Program Files (x86)\Yarn\bin\yarn.CMD
    npm: 8.5.2 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: Spartan (44.19041.1266.0), Chromium (105.0.1343.33)
    Internet Explorer: 11.0.19041.1566
  npmPackages:
    @sveltejs/adapter-auto: next => 1.0.0-next.71
    @sveltejs/adapter-node: next => 1.0.0-next.88
    @sveltejs/kit: next => 1.0.0-next.477
    svelte: ^3.48.0 => 3.50.0
    vite: ^3.1.0-beta.2 => 3.1.0
```

### Severity

annoyance

### LINKED ISSUE #518: Add option to auto-close inspector after a component has been opened
### Describe the problem

If you've used the inspector for any amount of time, you will have run into this:

1. Open Component using inspector -> Window switches to IDE
2. Do some editing 
3. Switch back to browser to see the result and immediately start clicking stuff, forgetting the inspector is still on -> Get sent back to the IDE by accident

This can be very frustrating, as it is quite easy to forget the inspector is still on.

Add an option to have the inspector auto-disable itself after opening something in the IDE. 

This behavior should not be the default, as it would make opening multiple components at once very tedious. But, people often only open one component at a time, so having an option to avoid being sent back to the IDE by accident would be nice.

### Alternatives considered

Simply stop forgetting the inspector is still on

### Importance

nice to have

### LINKED ISSUE #551: inspector: support vite base path for __openInEditor
### Describe the problem

see https://github.com/vitejs/vite/pull/11337

prepend base path if set

### Alternatives considered

leave as is, noone complained until now

### Importance

nice to have

dominikg:
We have received some Feedback for svelte-inspector and this PR aims to address most of it.

#### Adresses open issues and PRs:
fixes #434  by moving inspector ui outside of `<body>`
fixes #518  by automatically disabling inspector when user leaves the browser window after opening a file
fixes #551  by prepending configured base path to __openInEditor request
fixes #555  by changing default combo on linux to control-shift
fixes #559  by enabling holdMode by default

#### New Feature: Set options via environment.
```shell
# only set toggle
SVELTE_INSPECTOR_TOGGLE=control-shift

# or use complete config json
SVELTE_INSPECTOR_OPTIONS='{"holdMode": false}'
```

#### No longer experimental

Move your inspector config to the root of `vitePlugin`.
```diff
// svelte.config.js
export default {
  vitePlugin:{
    experimental:{
-      inspector: {...}
    },
+   inspector: {...}
  }
}
```

### DIFF
--- a/packages/vite-plugin-svelte/src/ui/inspector/plugin.ts
+++ b/packages/vite-plugin-svelte/src/ui/inspector/plugin.ts
@@ -41,6 +41,10 @@ export function svelteInspector(): Plugin {
 				...defaultInspectorOptions,
 				...options
 			};
+
+			inspectorOptions.__internal = {
+				base: config.base?.replace(/\/$/, '') || ''
+			};
 			const isSvelteKit = config.plugins.some((p) => p.name.startsWith('vite-plugin-sveltekit'));
 			if (isSvelteKit && !inspectorOptions.appendTo) {
 				// this could append twice if a user had a file that ends with /generated/root.svelte
--- a/packages/vite-plugin-svelte/src/ui/inspector/plugin.ts
+++ b/packages/vite-plugin-svelte/src/ui/inspector/plugin.ts
@@ -4,17 +4,7 @@ import path from 'path';
 import { fileURLToPath } from 'url';
 import fs from 'fs';
 import { idToFile } from './utils';
-import { InspectorOptions } from './options';
-
-const defaultInspectorOptions: InspectorOptions = {
-	toggleKeyCombo: process.platform === 'win32' ? 'control-shift' : 'meta-shift',
-	navKeys: { parent: 'ArrowUp', child: 'ArrowDown', next: 'ArrowRight', prev: 'ArrowLeft' },
-	openKey: 'Enter',
-	holdMode: false,
-	showToggleButton: 'active',
-	toggleButtonPos: 'top-right',
-	customStyles: true
-};
+import { defaultInspectorOptions, type InspectorOptions } from './options';
 
 function getInspectorPath() {
 	const pluginPath = normalizePath(path.dirname(fileURLToPath(import.meta.url)));
--- a/packages/vite-plugin-svelte/src/ui/inspector/plugin.ts
+++ b/packages/vite-plugin-svelte/src/ui/inspector/plugin.ts
@@ -4,7 +4,7 @@ import path from 'path';
 import { fileURLToPath } from 'url';
 import fs from 'fs';
 import { idToFile } from './utils';
-import { defaultInspectorOptions, type InspectorOptions } from './options';
+import { defaultInspectorOptions, type InspectorOptions, parseEnvironmentOptions } from './options';
 
 function getInspectorPath() {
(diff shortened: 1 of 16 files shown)


dominikg:
@delight sorry it took so long and i incorporated your PRs into this one as there was a fair bit of shifting around. I would love to get your review/feedback on this.

dominikg:
After discussing with svelte maintainers, inspector is not going to be enabled by default in vite-plugin-svelte as there can be situations where it is not useful and vite-plugin-svelte is a general purpose plugin.

It might be enabled by default in a future version of sveltekit though.

packages/vite-plugin-svelte/src/ui/inspector/plugin.ts
```diff
@@ -34,15 +24,38 @@ export function svelteInspector(): Plugin {
...
-				...options
+			if (config.command !== 'serve') {
+				log.debug(
+					`inspector only works for vite 'serve', disabled for '${config.command}'`,
+					undefined,
+					'inspector'
+				);
+				disabled = true;
+				return;
+			}
```
bluwy, line 47: The plugin already has `apply: 'serve'`, so I don't think this is needed.

bluwy (approved):
Trusting that you have tested this locally as I have not 😬 but the general changes LGTM.

PR #638, opened 2023-04-26
feat: remove SvelteKit specific handling

bluwy:
Per changeset:

Remove internal SvelteKit specific handling

- Disallow `kit` prop in inline options
- Remove default `hydratable: true` option for SvelteKit
- Inspector code mounts on `/@vite/client` to be framework agnostic

Re no1, I'm not sure how we're using that today, but seems safe to me to remove the runtime validation for it.

Re no2, SvelteKit has set it by itself for quite a while now: https://github.com/sveltejs/kit/blob/fb68c253a59402df2b2b1b5db7a54fad504cdd32/packages/kit/src/exports/vite/index.js#L136

Re no3, this removes the need for the `inspector.appendTo` option so we don't need kit specific handling for it. Makes way for https://github.com/sveltejs/vite-plugin-svelte/pull/631

### DIFF
--- a/.changeset/smooth-llamas-reflect.md
+++ b/.changeset/smooth-llamas-reflect.md
@@ -1,0 +1,9 @@
+---
+'@sveltejs/vite-plugin-svelte': minor
+---
+
+Remove internal SvelteKit specific handling
+
+- Disallow `kit` prop in inline options
+- Remove default `hydratable: true` option for SvelteKit
+- Inspector code mounts on `/@vite/client` to be framework agnostic
--- a/.changeset/smooth-llamas-reflect.md
+++ b/.changeset/smooth-llamas-reflect.md
@@ -3,7 +3,3 @@
 ---
 
 Remove internal SvelteKit specific handling
-
-- Disallow `kit` prop in inline options
-- Remove default `hydratable: true` option for SvelteKit
-- Inspector code mounts on `/@vite/client` to be framework agnostic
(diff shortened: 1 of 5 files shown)


packages/vite-plugin-svelte/src/ui/inspector/plugin.ts
```diff
@@ -84,32 +76,10 @@ export function svelteInspector(): Plugin {
 			}
 		},
 
-		transform(code: string, id: string, options?: { ssr?: boolean }) {
-			if (options?.ssr || disabled || !appendTo) {
-				return;
+		transform(code: string, id: string) {
+			if (id.includes('vite/dist/client/client.mjs')) {
+				return { code: `${code}\nimport('virtual:svelte-inspector-path:load-inspector.js')` };
```
bluwy, line 81: Dynamic import since the `Inspector.svelte` would rely on `/@vite/client` too, and it otherwise would cause an import loop an runtime `ReferenceError`.

dominikg:
the kit option and setting hydratable: true have been remnants for older versions of sveltekit. 
Does removing them make this a breaking change? 

Technically applications on older versions of kit can't update to newer v-p-s until they also update kit. But given that v-p-s is a dependency of kit and kit had a critical advisory that urged people to update to a version that is compatible, we could assume that updating v-p-s without also updating kit is not something we expect anyone to do?

dominikg (approved):

.changeset/smooth-llamas-reflect.md
```diff
@@ -0,0 +1,9 @@
+---
+'@sveltejs/vite-plugin-svelte': minor
+---
+
+Remove internal SvelteKit specific handling
+
+- Disallow `kit` prop in inline options
+- Remove default `hydratable: true` option for SvelteKit
+- Inspector code mounts on `/@vite/client` to be framework agnostic
```
dominikg, line 9:
```suggestion
Remove internal SvelteKit specific handling
```

PR #658, opened 2023-05-24
generate types with dts-buddy

Rich-Harris:
- types for exported functions are declared on the functions themselves, rather than in the `.d.ts` file
- a `types/index.d.ts` file is generated with an accompanying `types/index.d.ts.map`, meaning cmd-click takes you to the implementation

Unfortunately couldn't use the CLI for this because the entry point is a `.d.ts` file rather than a `.js` file, so this happens via a new `scripts/generate-dts.js` module


packages/vite-plugin-svelte/package.json
```diff
@@ -7,18 +7,18 @@
     "src"
```
dominikg, line 7: would `types` have to be added here ? 
```suggestion
    "src",
    "types"
```

packages/vite-plugin-svelte/package.json
```diff
@@ -7,18 +7,18 @@
...
+      "types": "./types/index.d.ts",
       "import": "./src/index.js"
     },
     "./package.json": "./package.json"
   },
   "scripts": {
     "dev": "pnpm typecheck --watch",
     "build": "pnpm typecheck",
-		"typecheck": "tsc --noEmit --allowJs --checkJs"
+    "typecheck": "node scripts/generate-dts.js"
```
dominikg, line 21: the script would have to be called something like build:types or generate:types to make it more clear that it has output. check suggests it doesn't change.

When exactly would this have to be called during release, how do you ensure that it's output in /types stays in sync with changes to types in /src

dominikg:
This basically duplicates src/public.d.ts into types/index.d.ts with the addition of a sourcemap. Is there a way to get the same result without the sourcemap and all these moving parts? v-p-s types are tiny, if all it takes is changing src/index.d.ts to declare the module in place instead of using declare function i'm fine with that.

packages/vite-plugin-svelte/package.json
Rich-Harris: elsewhere i'm using `prepublishOnly` for this

Rich-Harris:
> Is there a way to get the same result without the sourcemap and all these moving parts?

Unfortunately not, unless you want to create a sourcemap by hand. Though I think re-exporting from `index.js` (so the function types are declared inline rather than with `declare function`) will work in either case.

It _does_ add moving parts and the benefit is more modest in this case since there's really just one giant `.d.ts` file in `src` — no pressure to merge this PR, I was largely just curious about what it would look like (and I found a couple of `dts-buddy` bugs along the way, so it was worthwhile for me). Having cmd-click work properly _is_ nice, but might not be worth the trade-off. Your call

packages/vite-plugin-svelte/package.json
dominikg: that does not ensure it stays in sync. You would have to call it in `prepare`  (similar to how svelte called build in prepare). and even that isn't enough because types can still change and you end up with revisions where they are out of sync.

dominikg:
changed to draft for now until we figure out how to integrate it properly

dominikg:
closing in favor of https://github.com/sveltejs/vite-plugin-svelte/pull/751

PR #659, opened 2023-05-24
chore: cleanup after jsdoc

dominikg:
- remove unused dependencies
- improve rollup type use
- redo check scripts and include publint


.npmrc
```diff
@@ -0,0 +1,2 @@
+aggregate-output=true
+reporter=append-only
```
bluwy, line 2: Is there a reason these are added?

package.json
```diff
@@ -2,18 +2,21 @@
   "name": "vite-plugin-svelte-monorepo",
   "private": true,
   "scripts": {
-    "dev": "pnpm --dir packages/vite-plugin-svelte dev",
-    "build": "pnpm --dir packages/vite-plugin-svelte build",
+    "build": "echo 'no need to build'",
```
bluwy, line 5: It would be nice to remove this entirely since the npm/pnpm runner should also report something similar.

.npmrc
dominikg: output formatting for the check scripts

package.json
dominikg: done

PR #661, opened 2023-05-26
feat: add support for svelte-4 prereleases

dominikg:
This is to allow easier testing in the ecosystem. Until svelte 4.0.0 final is released, we are going to keep using svelte-3 in our main branch.


packages/vite-plugin-svelte-inspector/package.json
```diff
@@ -41,7 +41,7 @@
   },
   "peerDependencies": {
     "@sveltejs/vite-plugin-svelte": "^2.2.0",
-    "svelte": "^3.54.0",
+		"svelte": "^3.54.0 || ^4.0.0-next.0 || ^4.0.0",
```
benmccann, line 44: it looks like the rest of this file is using spaces, so this line is not aligned

benmccann (approved):
lgtm other than the minor indentation issue

PR #663, opened 2023-05-26
chore: enable prettier for json and reformat

dominikg:
can we add this to blame ignore here or do we have to merge into main first?

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -1,18 +1,18 @@
 {
-	extends: [
-		'config:base',
-		'schedule:weekly', // each monday
-		'group:allNonMajor', // batch non-majors into one PR
-		':semanticCommits' // nicer commit messages
+	"extends": [
+		"config:base",
+		"schedule:weekly", // each monday
+		"group:allNonMajor", // batch non-majors into one PR
+		":semanticCommits" // nicer commit messages
 	],
-	labels: ['dependencies'],
-	pin: false,
-	rangeStrategy: 'bump',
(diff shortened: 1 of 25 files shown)


ghostdevv:
All you should need is a file called `.git-blame-ignore-revs` with the commit hash in `7300d3b37b2b8b4099fccc0ecc3f8e4fef1151e8`

.github/renovate.json5
bluwy, line 1: This json5 formatting is a bit weird, but if you're fine with it, I'm fine with it too 😄

package.json
```diff
@@ -54,7 +54,7 @@
   "lint-staged": {
     "*.{js,ts,svelte,html,md,svx}": "eslint --cache --fix",
     "*.{css,scss,svelte,html,js,ts,svx,md}": [
-      "prettier --cache --write"
+      "prettier --cache --ignore-unknown --write"
     ]
```
bluwy, line 58: Should we expand this to json, json5, yml, mjs, etc? Or maybe a catch all `*`

bluwy:
> can we add this to blame ignore here or do we have to merge into main first?

We'd need to merge in main first since the commit hash will change after merge (commits are derived from commit time too)

.github/renovate.json5
dominikg: good catch, moved it to ignore

package.json
dominikg: thanks, * is correct together with --ignore-unknown.

PR #664, opened 2023-05-26
Mention editors for inspector

### LINKED ISSUE #647: docs: include documentation about how to use svelte inspector with custom editors
### Problem

It is not clear from the documentation how to use svelte-inspector with editors that are not supported by launch-editor (the library in vite that opens the editor for us) ootb.

### Solution

Add a link to launch-editor docs, mentioning `LAUNCH_EDITOR` environment variable

original report below:
---

### Describe the bug

There seems to be some detail missing, I know that these are "upstream" issues – but I think it's fair to say this plugin is about DX. So, with that in mind, I'd like to highlight some bugs/documentation.

1. The [documentation](https://github.com/sveltejs/vite-plugin-svelte/blob/main/docs/config.md#customizing-inspector-options-via-environment) states: 

> svelte inspector can be configured via environment variables.

Does that mean that I can specify: `VITE_INSPECTOR_EDITOR=code`, or `VITE_INSPECTOR_EDITOR=cursor`?

2. From the documentation, it seems a typical `svelte.config.js` might look like this:

```
const config = {
  vitePlugin: {
    exclude: [],

    // UX options:
    inspector: {
      toggleKeyCombo: 'meta-shift',
      holdMode: false,
      showToggleButton: 'never'
    }

  }

};

export default config;
```

But this leaves some details unanswered. Here's an example using VSC (Visual Studio Code):

- If you want to use this with VSC, then:
  - press `Cmd+Shift+P` (Mac) or `Ctrl+Shift+P` (Windows/Linux) to open the Command Palette. 
  - Then `Shell Command: Install 'code' command in PATH`.
  - Restart your terminal...
- By default, and behind the scenes this seems to do this: `npm run dev --openVim`, so you have to specify your editor in your terminal environment too, so:
  - in a terminal window, type: `echo $EDITOR`
  - if it's blank it'll be pointing to `vim`.
  - you can change it with: `export EDITOR=code`
  - then type: `echo $EDITOR` again, it should return `code`.
  - The Svelte inspector should work as intended.

3. what if you're on another IDE? – or perhaps you want a different editor to be saved?
  - open your terminal configuration (for example: ZSH):  `~/.zshrc`
  - then go to the bottom of the file and add the line: `export EDITOR=cursor`
  - restart yout terminal: `source ~/.zshrc`
  - `npm run dev -- --open`, and then `meta-shift` + click.
  - it should open in your saved IDE.

### Reproduction URL

https://kit.svelte.dev/

### Reproduction

The documentation says that this is all that needs configuring, but as described this is an incomplete solution:

```
const config = {
  vitePlugin: {
    exclude: [],

    // UX options:
    inspector: {
      toggleKeyCombo: 'meta-shift',
      holdMode: false,
      showToggleButton: 'never'
    }

  }

};

export default config;
```

### Logs

_No response_

### System Info

```shell
Apple Silicon, macOS Ventura, zsh, VSC or Cursor.so
```

ghostdevv:
Closes #647

### DIFF
--- a/docs/inspector.md
+++ b/docs/inspector.md
@@ -108,3 +108,7 @@ SVELTE_INSPECTOR_OPTIONS=true
   Inject custom styles when inspector is active. This is useful if you want to customize the inspector styles to match your app.
 
   When the inspector is active, the `svelte-inspector-enabled` class is added to the `body` element, and the `svelte-inspector-active-target` class is added to the current active target (e.g. via hover or keyboard).
+
+## Editors
+
+If your editor is not [supported out of the box](https://github.com/yyx990803/launch-editor#supported-editors), you can follow [the instructions here](https://github.com/yyx990803/launch-editor#custom-editor-support) to add it.


bluwy (approved):

dominikg:
thanks!

PR #675, opened 2023-06-22
chore: bump supported Svelte 4 version

bluwy:


dummdidumm:
AFAIK there's also some snapshots for Svelte 4 that need updating

bluwy:
I think dominik already handled that at https://github.com/sveltejs/vite-plugin-svelte/pull/661. Checking the ecosystem-ci logs, looks like it was failing because `version-4` branch doesn't exist in Svelte. (Ben fixed it but hasn't re-ran the tests)

dominikg (approved):

PR #680, opened 2023-06-26
chore: use svelte 4

bluwy:
close #678

- add svelte 3 in CI still, but only on ~~node16~~ node14 and ubuntu-latest
- remove `pnpmfile.cjs` as it's covered by `"vite": "$vite"` override
- Fix `semver` package vulnerability audit issue

~~Tests are intentionally failing until #679 is merged.~~ merged


.github/workflows/ci.yml
```diff
@@ -110,7 +118,7 @@ jobs:
         if: matrix.node != 14
         run: pnpm install --frozen-lockfile --prefer-offline --ignore-scripts
       - name: install for node14
-        if: matrix.node == 14
+        if: matrix.node == 14 || matrix.svelte == 3
```
dominikg, line 121: If svelte3 run finishes first, the pnpm cache saved might not work with svelte4. We may have to provide a custom cache key unless node action somehow figures it out. (eg by using a hash of the lockfile?)

package.json
```diff
@@ -62,7 +62,10 @@
   },
   "pnpm": {
     "overrides": {
-      "@sveltejs/vite-plugin-svelte": "workspace:^"
+      "@sveltejs/vite-plugin-svelte": "workspace:^",
+      "semver@<7.5.2": ">=7.5.2",
+      "svelte": "$svelte",
```
dominikg, line 67: What is this magic $? pnpm using svelte stores now?

.github/workflows/ci.yml
```diff
@@ -101,6 +106,9 @@ jobs:
         run: |
           npm i -g pnpm@^7.33.0
           tmppkg="$(jq '.engines.pnpm = "^7.33.0"' package.json)" && echo -E "${tmppkg}" > package.json && tmppkg=""
+      - name: use svelte 3
+        if: matrix.svelte == 3
+        run: cat package.json | jq '.devDependencies.svelte = "^3.59.2"' | tee package.json
```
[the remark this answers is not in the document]
bluwy: The strange thing with this is it seems to not work the first time, but only the second time 🤔 I copied this from https://github.com/svitejs/vitefu/blob/d5ee41e859ae8fabdb477e03e66dcb19b8a4636f/.github/workflows/ci.yml#L59 though

package.json
bluwy: Haha yeah pnpm has that documented at the bottom of [this section](https://pnpm.io/package_json#pnpmoverrides).

.github/workflows/ci.yml
bluwy: It doesn't seem like `setup-node` supports providing a custom cache key, but also I don't think mixing Svelte 3 and 4 would be an issue since it's only a cache with the same shape.

Worst case might be pnpm needing to download Svelte 3 and 4 from the network inconsistently, and that might be better than disabling cache for Svelte 3 altogether.

dominikg: hmm we can try it but I fear the cache gets busted in a way that is not good.
It would be better to check how to have separate caches for 3 and 4 ci

dominikg: The test failure archive needs svelte dimension in the name too

bluwy:
Looks like we can't run Svelte 4 in node14 CI since it doesn't support node14 anymore, so I made Svelte 3 to test in node14 instead. Don't think we need another CI run for Svelte 3 + node16? (which feels a bit wasteful)

dominikg (approved):

PR #692, opened 2023-07-20
fix: add svelte/internal/disclose-version  to optimizeDeps.include

dominikg:
it's used internally by svelte4, but vite dynamically discovering it leads to

```
9:33:40 PM [vite] ✨ new dependencies optimized: svelte/internal/disclose-version
9:33:40 PM [vite] ✨ optimized dependencies changed. reloading
```

so add it to our internal list thats used to generate optimizeDeps.include

### DIFF
--- a/packages/vite-plugin-svelte/src/utils/constants.js
+++ b/packages/vite-plugin-svelte/src/utils/constants.js
@@ -6,6 +6,7 @@ export const SVELTE_IMPORTS = [
 	'svelte/animate',
 	'svelte/easing',
 	'svelte/internal',
+	'svelte/internal/disclose-version',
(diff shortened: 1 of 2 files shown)


packages/vite-plugin-svelte/src/utils/constants.js
```diff
@@ -6,6 +6,7 @@ export const SVELTE_IMPORTS = [
 	'svelte/animate',
 	'svelte/easing',
 	'svelte/internal',
+	'svelte/internal/disclose-version',
```
bluwy, line 9: I think we need to optionally add this for Svelte 4 only.

bluwy (approved):

PR #700, opened 2023-07-28
fix: use full filename for svelte.compile 

### LINKED ISSUE #699: link to correct source in error page
### Describe the bug

Currently the error pages do not link through to the correct page

Screenshot 2023-07-27 at 11 40 22

the link is just relative to the project "/src/lib/components/ProjectMain.svelte:127:17" - but it needs the but it needs the file:// prefix plus the location of the project.

I know it must be possible since with the svelte inspector it's possible to open the relevant component with a click.

### Reproduction URL

https://github.com

### Reproduction

I didn't create a repro since it happens on all repos

### Logs

_No response_

### System Info

```shell
Chrome, MacOs, VSCode
```

dominikg:
fixes #699 

There is a small chance that this creates a regression with build output still containing absolute paths in sourcemap sources, but in my tests with most recent svelte and vite versions, that is no longer the case.


bluwy (approved):
If this produces stable builds on different machines, then it sounds like the best and simplest fix for me 👍

dominikg:
looks like the windows test is failing, ran it in my local windows 10 vm and it produced a development build with `add_location` calls? wtf

dominikg:
turns out something in our e2e tests sets `process.env.NODE_ENV` to `TEST` which caused viteConfig.isProduction to be false, which in turn enabled the svelte compiler option `dev`. 

A better default value for that would be `command==='serve' || mode==='development' || process.env.NODE_ENV ==='development'`  but that is a breaking change, so for now we force production build in our tests via NODE_ENV=production

bluwy (approved):

PR #747, opened 2023-09-15
refactor(resolve): prefer exports condition over svelte field

### LINKED ISSUE #725: prefer svelte exports condition over svelte field for resolving libraries
### Describe the problem

We want to remove support for svelte field altogether at some point, take gradual steps to ensure existing libraries can update to not hit users with errors.

Start prefering the exports condition value over svelte field, add a more persistent/pressing warning for libraries that only use svelte field to switch to exports.

In a following major, drop svelte field resolve entirely. (possibly together with adding svelte 5 support)

### Alternatives considered

give it more time.

### Importance

would make my life easier

dominikg:
This is the second step in the migration to exports condition.

Previously, the svelte field took precedence, now it is only used as a fallback if vite is unable to resolve a library.

The warning log and faq have been updated to reflect that library authors need to add the exports condition to remain compatible with future majors.

fixes #725


dominikg:
@bluwy  noticed that if we prefer exports condition, we can actually rely on vite resolving the svelte field via mainFields.

so we are able to get rid of the custom resolve logic completely. yay

bluwy (approved):
🚢

PR #751, opened 2023-09-16
refactor(types): use dts-buddy to generate types with map

dominikg:
and cleanup type definitions and tsconfig

This introduces a `types:generate` script that has to be called every time you change a type to update types/index.d.ts which is commited.

Having the generated file in the repo ensures that we are able to review changes in public types and can prevent unintended drift or breaking changes


dominikg:
@Rich-Harris  in #658  you did an export of index.js from a d.ts files which I didn't use here. Was that something that has changed in dts-buddy or is this one here holding it wrong?

dominikg:
example of the script that validates types are in sync https://github.com/sveltejs/vite-plugin-svelte/actions/runs/6207076288/job/16851955290?pr=751#step:12:18

dominikg:
hmm, dts-buddy pulling in an older version of ts isn't optimal...

Rich-Harris:
Bumped dts-buddy, so that it now consumes TypeScript as a peer dependency.

> in https://github.com/sveltejs/vite-plugin-svelte/pull/658 you did an export of index.js from a d.ts files which I didn't use here

The reason for that is that people should be able to do this...

```ts
import type { CompileOptions } from '@sveltejs/vite-plugin-svelte';
```

...which isn't possible if the `index.js` is the entry module, since it doesn't (can't) re-export the stuff in `public.d.ts`. In cases where we want to expose both types and values, but are authoring in `.js` with JSDoc, we need to treat the `.d.ts` file as the entry point and re-export the values from the _real_ entry point.

dominikg:
hmm, i think this works. Not for CompileOptions as that is now imported from svelte-4 and only part of the Options type of vite-plugin-svelte, but here is the type in the namespace
https://github.com/sveltejs/vite-plugin-svelte/pull/751/files#diff-3ce3450c71e77215bb00652cd1afec4105620695f279d2c142552a181c3fb877R7

and at least in my local tests `import type {Options} from '@sveltejs/vite-plugin-svelte'`  works.  (for the rare case a user needs to do it at all,  usually they just do 
```ts
import {svelte} from '@sveltejs/vite-plugin-svelte';
import {defineConfig} from 'vite';
export default defineConfig({
  plugins: [
    svelte({
      // yay for autocomplete
    })
  ]
})
```

note that in this repo the output of dts-buddy is commited, so /types/index.d.ts is what is going to get published. a mismatch is prevented by calling a script that ensures the output is in sync with the current repo content.

.changeset/metal-olives-wait.md
```diff
@@ -0,0 +1,6 @@
+---
+'@sveltejs/vite-plugin-svelte-inspector': major
+'@sveltejs/vite-plugin-svelte': major
+---
+
+breaking(types): emit types with dts-buddy to include type map"
```
bluwy, line 6:
```suggestion
breaking(types): emit types with dts-buddy to include type map
```

There's a lone double quote

packages/vite-plugin-svelte-inspector/types/index.d.ts
```diff
@@ -0,0 +1,74 @@
...
+		 */
+		customStyles?: boolean;
+
+		/**
+		 * internal options that are automatically set, not to be set or used by users
+		 * */
+		__internal?: {
+			// vite base url
+			base: string;
+		};
```
bluwy, line 69: This isn't marked `@internal` by dts-buddy. Perhaps not a big deal but figured to flag it here. Original:

https://github.com/sveltejs/vite-plugin-svelte/blob/24a14e4d6fe0758d619442cc62efcb24416c1bd4/packages/vite-plugin-svelte-inspector/src/public.d.ts#L62-L69

bluwy (commented):
Still not a fan of dts-buddy 😬 But I won't block the PR if you want to merge it.

packages/vite-plugin-svelte-inspector/types/index.d.ts
dominikg: noted https://github.com/Rich-Harris/dts-buddy/issues/63

Issue #733, opened 2023-09-04
vite-plugin-svelte chokes on custom media syntax when using vitePreprocess and Lightning CSS


knpwrs:
### Describe the bug

I just spun up a new skeleton sveltekit project which by default uses `vitePreprocess`. I installed lightning css (`npm i -D lightningcss`) I have modified `vite.config.ts` to contain the following:

```ts
import { sveltekit } from '@sveltejs/kit/vite';
import { defineConfig } from 'vite';

export default defineConfig({
  plugins: [sveltekit()],
  css: {
    transformer: 'lightningcss',
    lightningcss: {
      drafts: {
        nesting: true,
        customMedia: true,
      },
    },
  },
});
```

Given the following `+page.svelte` file, I can see the `nesting` configuration works:

```html
<h1>Welcome to SvelteKit</h1>

<style>
  /* @custom-media --small-viewport (max-width: 30em); */

  h1 {
    color: blue;

    &:hover {
      color: red;
    }

    /* @media (--small-viewport) { */
    /*   color: green; */
    /* } */
  }

</style>
```

With the configuration set as in my first code block, hovering the `<h1>` turns it red. If I comment out the `nesting` configuration, hovering the `<h1>` does nothing.

However, if I uncomment the custom media, I get a syntax error:

```
Internal server error: /Users/knpwrs/Downloads/my-app/src/routes/+page.svelte:4:43 ")" is expected
  Plugin: vite-plugin-svelte
  File: .../my-app/src/routes/+page.svelte:4:43
   2 |
   3 |  <style>
   4 |    @custom-media --small-viewport (max-width: 30em);
                                                   ^
```

### Reproduction URL

https://github.com/knpwrs/sveltekit-lightningcss-custom-media-issue

### Reproduction

1. `npm create svelte@latest my-app`
2. `cd my-app`
3. `npm install -D lightningcss`
4. Update configuration as shown above
5. Observe bug

### Logs

[logs.txt](https://github.com/sveltejs/vite-plugin-svelte/files/12513452/logs.txt)

### System Info

```shell
System:
    OS: macOS 13.4
    CPU: (12) arm64 Apple M2 Max
    Memory: 1.26 GB / 64.00 GB
    Shell: 5.9 - /opt/homebrew/bin/zsh
  Binaries:
    Node: 20.3.1 - /opt/homebrew/bin/node
    Yarn: 1.22.19 - /opt/homebrew/bin/yarn
[...]

PR #756, opened 2023-09-23
feat!: preprocess style tags by default

bluwy:
fix https://github.com/sveltejs/vite-plugin-svelte/issues/733

Follow Vite's behavior to preprocess all CSS by default with postcss or lightningcss.

Note: there's a small optimization in Vite where it skips postcss if there's no `@import` and postcss plugins. Lightningcss doesn't seem to have that optimization.

This affects preprocessing in component libraries too, but I think they should only have a small impact.


kaisermann (approved):
I guess this is fine too! Repeating what I said in the linked issue: `svelte-preprocess` [will preprocess style blocks without `lang` if the `postcss` configuration is found/valid.](https://github.com/sveltejs/svelte-preprocess/blob/main/src/autoProcess.ts#L237).

benmccann:
My main question is whether language-tools would know to run the preprocessor, so I'd want Simon's approval

benmccann:
@dummdidumm is on vacation, so you might have to ask again later 😄

dummdidumm (approved):
I'm good with this. The fact that the IDE may give false positives shouldn't hold up this valuable PR - and people can tune their config / add a lang tag if needed. We may also be able to detect some of this inside language tools and don't show CSS errors. And since nested CSS is now part of the spec, the upstream vs code html language service might even support it at some point.

dominikg (approved):

dominikg:
If people report performance regressions from postcss going all over their files without lang=postcss tag, we can add an option to restore the previous behavior or recommend they switch to lightningcss?

Issue #585, opened 2023-01-31
Provide helpful error when using `lang="ts"` without a preprocessor


Rich-Harris:
### Describe the problem

see e.g. https://github.com/sveltejs/kit/pull/8811. If a developer adds `lang="ts"` to a component without the relevant preprocessor, they'll get 'unexpected token' errors.

### Describe the proposed solution

This is probably a common enough case that we could intercept the error and say something like 'you should add `preprocessor: vitePreprocess()`' or whatever

### Alternatives considered

continue as presently

### Importance

nice to have

dominikg:
Question is how do you detect that without adding overhead for everyone.

The only way i can think of rn is using `toRollupError` (which basically wraps compile anyways to make vite error overlay nicer) and in there do
```
if( token_error && lang_ts && preprocessor_missing) {
  rollupError.message = 'friendly reminder to install ts preprocessor' 
}
```

problem is that lang_ts needs a regex and preprocessor_missing is hard to determine. Only really safe way is to feed a dummy ts script into the preprocessor chain and see if that fails.

Which reads like a lot of complexity. How often has this problem been reported so far?

yfwz100:
I also came across this error and was wondering why typescript is not supported in a modern framework... If it's hard to detect, can we add this pre-processor by default?

dominikg:
> I also came across this error and was wondering why typescript is not supported in a modern framework... If it's hard to detect, can we add this pre-processor by default?

vite-plugin-svelte itself has some automatic configuration for svelte features that are required for vite hot module reloading to work.
However it won't start adding opinionated configuration for optional features. 

It's up to the starter templates/frameworks/users that use typescript to add the required preprocessors and configuration needed for their projects. 

Best we can do is try to have a better error message when we detect that it is missing.

rnwolfe:
I am getting this error _with_ `vitePreprocess()` setup.

## `svelte.config.js`:
```js
import adapter from '@sveltejs/adapter-node';
import { vitePreprocess } from '@sveltejs/kit/vite';

/** @type {import('@sveltejs/kit').Config} */
const config = {
	// Consult https://kit.svelte.dev/docs/integrations#preprocessors
	// for more information about preprocessors
	preprocess: vitePreprocess(),
	kit: {
		adapter: adapter(),
		csp: {
			directives: {
				'font-src': ['self', 'data:', '*.typekit.net', '*.cloudflare.com'],
				'script-src': ['self', '*']
			}
		}
[...]

PR #757, opened 2023-09-23
feat: improve compile error

bluwy:
Fix https://github.com/sveltejs/vite-plugin-svelte/issues/585

image

image

Also added names for our preprocessors. Not very fond that we now have a lot of logic for error handling, but probably still good for users in the long run (as long as the logic itself dont cause more errors)

### DIFF
--- a/packages/vite-plugin-svelte/src/utils/error.js
+++ b/packages/vite-plugin-svelte/src/utils/error.js
@@ -125,13 +125,15 @@ export function enhanceCompileError(err, originalCode, preprocessors) {
 			const isErrorInScript = matchStart <= errIndex && errIndex <= matchEnd;
 			if (isErrorInScript) {
 				// Warn missing lang="ts"
-				if (!m[1]?.includes('lang="ts"')) {
-					additionalMessages.push('Did you forgot to add lang="ts" to your script tag?');
+				const hasLangTs = m[1]?.includes('lang="ts"');
+				if (!hasLangTs) {
+					additionalMessages.push('Did you forget to add lang="ts" to your script tag?');
 				}
 				// Warn missing script preprocessor
 				if (preprocessors.every((p) => p.script == null)) {
+					const preprocessorType = hasLangTs ? 'TypeScript' : 'script';
 					additionalMessages.push(
-						'Did you forgot to add a script preprocessor? See https://github.com/sveltejs/vite-plugin-svelte/blob/main/docs/preprocess.md for more information.'
+						`Did you forget to add a ${preprocessorType} preprocessor? See https://github.com/sveltejs/vite-plugin-svelte/blob/main/docs/preprocess.md for more information.`
 					);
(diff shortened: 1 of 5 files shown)


packages/vite-plugin-svelte/src/utils/error.js
```diff
@@ -100,3 +100,76 @@ function formatFrameForVite(frame) {
...
+
+		let m;
+		while ((m = scriptRe.exec(originalCode))) {
+			const matchStart = m.index;
+			const matchEnd = matchStart + m[0].length;
+			const isErrorInScript = matchStart <= errIndex && errIndex <= matchEnd;
+			if (isErrorInScript) {
+				// Warn missing lang="ts"
+				if (!m[1]?.includes('lang="ts"')) {
+					additionalMessages.push('Did you forgot to add lang="ts" to your script tag?');
```
dominikg, line 129: "Did you forget..." (for some reason i can't add suggestions here)

packages/vite-plugin-svelte/src/utils/error.js
```diff
@@ -100,3 +100,76 @@ function formatFrameForVite(frame) {
...
+				}
+				// Warn missing script preprocessor
+				if (preprocessors.every((p) => p.script == null)) {
+					additionalMessages.push(
+						'Did you forgot to add a script preprocessor? See https://github.com/sveltejs/vite-plugin-svelte/blob/main/docs/preprocess.md for more information.'
```
dominikg, line 134: this can be a bit more specific for a typescript preprocessor if we detect lang=ts is present, and else we can mention the lang value (what others are there? coffeescript? :D

packages/vite-plugin-svelte/src/utils/error.js
```diff
@@ -100,3 +100,76 @@ function formatFrameForVite(frame) {
...
+			// Warn missing lang attribute
+			if (!m[1]?.includes('lang=')) {
+				additionalMessages.push('Did you forgot to add a lang attribute to your style tag?');
+			}
+			// Warn missing script preprocessor
+			if (
+				preprocessors.every((p) => p.style == null || p.name === 'inject-scope-everything-rule')
+			) {
+				additionalMessages.push(
+					'Did you forgot to add a style preprocessor? See https://github.com/sveltejs/vite-plugin-svelte/blob/main/docs/preprocess.md for more information.'
```
dominikg, line 156: same here, if we know the lang attribute, it would be nice to include it.

packages/vite-plugin-svelte/src/utils/error.js
bluwy: I'm focusing on TypeScript only for now as that's the main one Vite supports, and IIRC CoffeeScript doesn't really work in Vite. So TypeScript has a little special treatment here.

dominikg (approved):

PR #760, opened 2023-09-30
experiment: try to declare virtual css modules sideeffect free

dominikg:
unfortunately the only thing that seems to work is using the `sideEffects` field in package.json.

using annotation comments isn't working for import/export and both returning moduleSideEffects: false from the load hook or a custom rollup.treeshake function does not lead to the magenta css being omitted in the testcase :(

### DIFF
--- a/packages/e2e-tests/css-nosideeffects/__tests__/css-nosideeffects.spec.ts
+++ b/packages/e2e-tests/css-nosideeffects/__tests__/css-nosideeffects.spec.ts
@@ -1,0 +1,20 @@
+import { browserLogs, findAssetFile, getColor, getText, isBuild } from '~utils';
+
+test('should not have failed requests', async () => {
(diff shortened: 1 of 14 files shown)


bluwy:
Looks like it's being overridden at  https://github.com/vitejs/vite/blob/22bd67d70a1390daae19ca33d7de162140d533d6/packages/vite/src/node/plugins/css.ts#L515

If I add a post Vite plugin to force it to `moduleSideEffects: false`, the CSS are all gone as Vite is unable to retrieve them in `renderChunk` (as the linked code comment showed). The `package.json` `sideEffects` trick worked because it sets `*.svelte` files as no side-effects instead, which covers the CSS treeshaking one level higher, so it doesn't hit the Vite caveat.

This seems a little tricky. Maybe Vite needs to handle CSS `moduleSideEffects: false` specifically.

PR #764, opened 2023-10-05
chore: use type module in root

bluwy:
Had to also tweak some strange eslint linting with this change.

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -92,6 +92,18 @@ module.exports = {
 			}
 		},
 		{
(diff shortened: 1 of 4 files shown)


dominikg:
should we drop support for svelte.config.cjs?

dominikg:
users can add a proxy svelte.config.js that loads it with createRequire if they really insist

bluwy:
probably a bit of a stretch i think, maybe we can start with a warning if we want to though

dominikg (approved):

dominikg (approved):

PR #765, opened 2023-10-06
feat: promote experimental.dynamicCompileOptions

dominikg:
The option has been there for a while and is used. Promote it to proper with version 3.0.

Technically not a breaking change as we copy it over and log a warning.


bluwy (approved):
Fine with keeping the older promoted options for now 👍

Issue #763, opened 2023-10-04
Circular filesystem watch events cascade, including self-referential, leading to RangeError: Maximum call stack size exceeded


andriytyurnikov:
### Describe the bug

I have encountered crash of dev script.
```
RangeError: Maximum call stack size exceeded
```
Initial investigation lead me to `emitChangeEventOnDependants` function in `src/utils/watch.js`, where I've discovered 
self-referential loops, where supposed `dependant` was same as `filename`.

Unfortunately, blocking self-referential event emitting revealed another problem - circular emission of virtual change event.

It is challenging to create an isolated reproduction. 
Which may or may not be related to how vite-plugin-svelte-cache choses to build dependency tree. 

However, it is clear  that self-referential, and circular references may exist, and therefore may overwhelm nodejs process with storm of emited events.

While it is not as good as reproduction, I hope, you would find this information instrumental.

### Reproduction URL

https://github.com/sveltejs/vite-plugin-svelte

### System Info

```shell
System:
    OS: macOS 13.5.2
    CPU: (8) arm64 Apple M2
    Memory: 208.14 MB / 24.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.8.0 - ~/.asdf/installs/nodejs/20.8.0/bin/node
    npm: 10.1.0 - ~/.asdf/plugins/nodejs/shims/npm
  Browsers:
    Chrome: 117.0.5938.149
    Safari: 16.6
  npmPackages:
    @sveltejs/adapter-auto: ^2.0.0 => 2.1.0 
    @sveltejs/adapter-static: ^2.0.3 => 2.0.3 
    @sveltejs/kit: ^1.20.4 => 1.25.1 
    svelte: ^4.0.5 => 4.2.1 
    vite: ^4.4.2 => 4.4.10
```

bluwy:
Related: https://github.com/sveltejs/vite-plugin-svelte/issues/690

dominikg:
please provide a reproduction, I'm eager to kearn where this happens. circular dependencies in preprocess dependencies is always an error i think. we can add a guard to avoid the stack error, but would throw a "circular dependency" error instead.

andriytyurnikov:
@dominikg, i get it by commenting and uncommenting styles multiple times in https://github.com/andriytyurnikov/andriytyurnikov.github.io/src/routes/garage/web-font-size-calculator/+page.svelte

[...]

PR #767, opened 2023-10-15
feat(preprocess): check preprocess dependencies for anomalies

dominikg:
#### depending on itself 
this is usually a sign for a bad setup/import, see #763 

#### depending on many files
all these files have to be read separately and changing one of them leads to a hot reload of the component. Depending on a few files is ok, but many cause issues.

Messages get added as extra compiler warnings using the `vite-plugin-svelte-preprocess-` prefix. This way users can opt to ignore them via `onwarn` hook. 

Are these checks too rigid and do we need FAQ entries and link them in the messages?

### DIFF
--- a/packages/vite-plugin-svelte/src/utils/preprocess.js
+++ b/packages/vite-plugin-svelte/src/utils/preprocess.js
@@ -130,13 +130,23 @@ export function addExtraPreprocessors(options, config) {
  * @returns {({dependencies: string[], warnings:import('svelte/types/compiler/interfaces').Warning[] })}
  */
 export function checkPreprocessDependencies(filename, dependencies) {
-	const normalizedFullFilename = normalizePath(filename);
 	/** @type {import('svelte/types/compiler/interfaces').Warning[]} */
 	const warnings = [];
-	const filteredDeps = dependencies
-		.map(normalizePath)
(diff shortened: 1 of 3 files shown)


packages/vite-plugin-svelte/src/utils/preprocess.js
```diff
@@ -121,3 +122,52 @@ export function addExtraPreprocessors(options, config) {
...
+		if (normalizedDeps[i] === normalizedFullFilename) {
+			selfIdx.push(i);
+		}
+	}
+	const hasSelfDependency = selfIdx.length > 0;
+	if (hasSelfDependency) {
+		warnings.push({
+			code: 'vite-plugin-svelte-preprocess-depends-on-self',
+			message:
+				'svelte.preprocess returned this file as a dependency of itself. This can be caused by invalid configuration or importing generated code that depends on .svelte files (eg. tailwind base css)',
```
bluwy, line 153:
```suggestion
				'svelte.preprocess returned this file as a dependency of itself. This can be caused by an invalid configuration or importing generated code that depends on .svelte files (eg. tailwind base css)',
```

bluwy (approved):
Some message nits

PR #780, opened 2023-11-03
chore: update tsconfig setup

bluwy:
Vite 5 re-exports `rollup/parseAst` which is only supported by TS if it respects the `"exports"` field. I updated the tsconfigs so it works now.


packages/vite-plugin-svelte/tsconfig.json
```diff
@@ -3,8 +3,8 @@
 	"compilerOptions": {
 		"noEmit": true,
 		"target": "ES2020",
```
dominikg, line 5: can we update target here?

bluwy: You mean something higher like "es2022"? It doesn't support "node*" strings

dominikg (approved):

PR #785, opened 2023-11-09
generate: 'client' | 'server'

Rich-Harris:


Rich-Harris:
hmm, not sure how to deal with the types... might have to just ignore them

dominikg (approved):

dominikg:
went ahead and ignored them for now.

would this become a problem for anyone else who wants to support 4 and 5? ( rollup-plugin-svelte, svelte-loader etc).

cc @dummdidumm  maybe you have an idea how to solve it (or is it ok to just ignore?).

dummdidumm:
I'd say just slap an ignore there. If you know what you're doing as a tooling author it's fine

PR #787, opened 2023-11-10
chore: cleanup

dominikg:


packages/vite-plugin-svelte/src/utils/compile.js
```diff
@@ -203,7 +205,7 @@ function buildMakeHot(options) {
 		// @ts-ignore
 		const adapter = options?.hot?.adapter;
 		return createMakeHot({
-			walk,
+			walk: svelte.walk,
```
bluwy, line 208: Should we also put a TODO here like in the v2 branch:

https://github.com/sveltejs/vite-plugin-svelte/blob/ec4c530125b446e1d8b5f5f84e62aafee6275320/packages/vite-plugin-svelte/src/utils/compile.js#L191-L192

packages/vite-plugin-svelte/src/utils/constants.js
```diff
@@ -6,14 +8,16 @@ export const SVELTE_IMPORTS = [
...
 	'svelte/ssr',
 	'svelte/store',
 	'svelte/transition',
 	'svelte'
 ];
 
+if (!isSvelte5) {
+	// TODO add back to list above once it's part of svelte5
+	SVELTE_IMPORTS.push('svelte/internal/disclose-version', 'svelte/motion');
+}
```
bluwy, line 20: It seems like the v2 branch does this instead:

https://github.com/sveltejs/vite-plugin-svelte/blob/ec4c530125b446e1d8b5f5f84e62aafee6275320/packages/vite-plugin-svelte/src/utils/constants.js#L17-L20

Which I think means we can revert this part and put it in `SVELTE_IMPORTS` always.

packages/vite-plugin-svelte/src/index.js
```diff
@@ -185,9 +189,43 @@ export function svelte(inlineOptions) {
...
+		}
 	];
+	if (!isSvelte5) {
+		plugins.push(svelteInspector()); // TODO reenable once svelte5 has support
+	}
+	if (isSvelte5) {
+		log.warn(
+			'svelte 5 support in vite-plugin-svelte is experimental, breaking changes can occur in any release until this notice is removed'
+		);
+		log.warn('svelte5 does not support svelte-inspector yet, disabling it');
```
bluwy, line 201:
```suggestion
		log.warn('svelte 5 does not support svelte-inspector yet, disabling it');
```

packages/vite-plugin-svelte/src/utils/compile.js
dominikg: no, this won't ever be used for 5, hmr is going to work differently, likely with a compiler option.

packages/vite-plugin-svelte/src/utils/constants.js
dominikg: yes, disclose version has been added to 5 after this

packages/vite-plugin-svelte/src/index.js
dominikg: refactored so it's a single log and changed wording a bit

bluwy (approved):
Nice cleanup

PR #796, opened 2023-11-16
fix: handle closing tags in attributes

dummdidumm:
...when detecting script language
Related to https://github.com/sveltejs/language-tools/issues/2204
Let me know if there's any other place where this needs a fix

### DIFF
--- a/packages/vite-plugin-svelte/src/utils/compile.js
+++ b/packages/vite-plugin-svelte/src/utils/compile.js
@@ -7,7 +7,8 @@ import { log } from './log.js';
 import { createInjectScopeEverythingRulePreprocessorGroup } from './preprocess.js';
 import { mapToRelative } from './sourcemaps.js';
 
-const scriptLangRE = /<script [^>]*lang=["']?([^"' >]+)["']?[^>]*>/;
+const scriptLangRE =
+	/<script (?:[^>]*|(?:[^=>'"/]+=(?:"[^"]*"|'[^']*'|[^>\s])\s+)*)lang=["']?([^"' >]+)["']?[^>]*>/;
 
 import { isSvelte3 } from './svelte-version.js';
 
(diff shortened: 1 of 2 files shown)


bluwy (approved):
LGTM.

If we backport this to `main`, we also need to update these two places (for error reporting only):

https://github.com/sveltejs/vite-plugin-svelte/blob/cf1ef58ff0fd21bd5de1216eed2e1d378e88fc35/packages/vite-plugin-svelte/src/utils/error.js#L118

https://github.com/sveltejs/vite-plugin-svelte/blob/cf1ef58ff0fd21bd5de1216eed2e1d378e88fc35/packages/vite-plugin-svelte/src/utils/error.js#L145

dummdidumm:
Thanks! I'll do a corresponding PR for main 👍

dominikg:
can we do it on main first and cherry pick the change to v2? now that main is no longer in pre-release that flow makes the most sense to me. although with the amount of changes between main and v2 it can become difficult to pick sometimes.

packages/vite-plugin-svelte/src/utils/compile.js
```diff
@@ -7,7 +7,8 @@ import { log } from './log.js';
 import { createInjectScopeEverythingRulePreprocessorGroup } from './preprocess.js';
 import { mapToRelative } from './sourcemaps.js';
 
-const scriptLangRE = /<script [^>]*lang=["']?([^"' >]+)["']?[^>]*>/;
+const scriptLangRE =
+	/<script (?:[^>]*|(?:[^=>'"/]+=(?:"[^"]*"|'[^']*'|[^>\s])\s+)*)lang=["']?([^"' >]+)["']?[^>]*>/;
```
dominikg, line 11:
```suggestion
	/<script (?:[^>]*|(?:[^=>'"/]+=(?:"[^"]*"|'[^']*'|[^>\s]+)\s+)*)lang=["']?([^"' >]+)["']?[^>]*>/;
```

@dummdidumm following the fixes in svelte, this has to be updated too i think

bluwy: Now that I think about it. The chances of someone doing `<script generic="T extends Record<string, any>" lang="ts"` is very low that I won't mind keeping it simple as before.

But I also don't mind the more comprehensive check. Just noting it here.

PR #797, opened 2023-11-17
Fix optional parameter types

bluwy:
The parameter types are optional but since https://github.com/sveltejs/vite-plugin-svelte/pull/751 it was accidentally marked required.


bluwy:
CI is failing because `generate:types` generates `import("vite")` but the formatter changes to `import('vite')` on commit 🙃  But we don't run the formatter before validating the types.

dominikg (approved):

dominikg (approved):

PR #801, opened 2023-11-18
Update faq.md : Unknown file extension ".svelte" : how to solve

improveTheFuture:
Too few help for this issue on the web...

Solution 1: Found here https://stackoverflow.com/a/76570669
Solution 2: Found here https://githubissues.com/hperrin/svelte-material-ui/608#error
But FORGOT TO MENTION to "just stop the running vite dev server (if any), and run ```npm run dev``` again". Instead of simply "just run ```npm run dev```."
**Could you edit the commit with this precision before merging ?**

Solution 3: Worked for me

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -236,3 +236,9 @@ See the faq about [vite and prebundling](#what-is-going-on-with-vite-and-pre-bun
 > [svelte-package](https://kit.svelte.dev/docs/packaging) which already supports it.
 >
 > For backwards compatibility, you can keep the `svelte` field in addition to the `exports` condition. But make sure that both always resolve to the same files.
+
+# Unknown file extension ".svelte"
+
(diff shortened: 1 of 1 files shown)


dominikg:
Thank you for this suggestion. To learn more about your problem please share a reproduction. The links you provided describe different problems. The first one looks like a sveltekit issue where the manifest is not up to date, it's out of scope for vite-plugin-svelte.

Please focus on the second one that describes a problem when installing new dependencies while the devserver is running. Vite itself is able to detect new dependencies and trigger a new optimization+reload at runtime so there could be a bug here that needs to be fixed. So the right course of action here is to file an issue with a reproduction. If it turns out that restarting the dev server is indeed the recommended resolution for this, we can work on improving the FAQ towards this.

PR #802, opened 2023-11-18
Update faq.md : missing exports condition : svelte property must be r…

improveTheFuture:
…emoved

If the svelte property remains in package.json, running "npm run dev" still displays the warning :  "[vite-plugin-svelte] WARNING: The following packages use a svelte resolve configuration in package.json that has conflicting results and is going to cause problems future."

Confirmed by the [SvelteKit Packaging Doc](https://kit.svelte.dev/docs/packaging#anatomy-of-a-package-json-svelte) that mentioned it as being only a field.

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -221,7 +221,7 @@ Example:
 ```diff
 // package.json
   "files": ["dist"],
-  "svelte": "dist/index.js",
+-  "svelte": "dist/index.js",
 + "exports": {
 +   ".": {
 +     "svelte": "./dist/index.js"


dominikg:
Thank you for working on improving our documentation!

For the time being, we don't want to recommend removing the top level svelte field entirely, the faq entry mentions this at the bottom:

> For backwards compatibility, you can keep the svelte field in addition to the exports condition. But make sure that both always resolve to the same files.

The same is mentioned by the packaging faq you linked

> but for backwards compatibility with outdated tooling that doesn't yet know about export conditions it's good to keep it around. It should point towards your root entry point.

So the FAQ entry is based on current recommended best practices.

I'm going to close this PR but please open a bug report with a reproduction if you have a library that displays the warning you mentioned despite following these practices. You can also ask for help on https://svelte.dev/chat in `#help-svelte-and-kit` there are many active members who publish and use libraries from npm and are able to help.

Issue #803, opened 2023-11-19
Suppress "ignoring compiler output js change..." message unless debugging enabled


Radiergummi:
### Describe the problem

If code in a component doesn't functionally change, but has updates in whitespace after saving, the plugin will emit a message such as this:
```
10:50:22 AM [vite-plugin-svelte] ignoring compiler output js change for /path/to/component.svelte as it
is equal to previous output after normalisation
```

This happens frequently when you reformat your code, perhaps even automatically in your IDE, so the console will contain a lot of these.  
I appreciate tooling letting me know what happens, but this message is neither actionable nor relevant. Is there any reason to emit it at all, unless debugging is enabled?

### Describe the proposed solution

I would like to suppress this log output, unless verbose or debugging mode is explicitly enabled.

### Alternatives considered

I know this issue is a nitpick, but improving DX is always good :)

### Importance

nice to have

dominikg:
This is a reasonable request, i think we can lower the loglevel of this to debug. Would you want to work on a PR implementing it? Should be very easy (find the log, change info to debug and add a changeset)

PR #806, opened 2023-11-20
Lowers log level to debug

Radiergummi:
This message isn't immediately actionable for developers and can reasonably be considered debugging output. Therefore, the log level has been updated to reflect this.

This fixes #803.

PR #807, opened 2023-11-20
log level changed from warn to debug

Augasty:
Changed the log level for whitespace update from warn to debug.

This solves #803


dominikg:
closing in favor of #806

Issue #808, opened 2023-11-21
Default Vite mainFields is incorrect (missing "browser")


WoodyWoodsta:
### Describe the bug

With Vite 5, `browserField` is [removed](https://github.com/vitejs/vite/pull/14733), requiring `"browser"` to be included in the config `resolve.mainFields` to ensure correct resolving of browser exports.

The default value for this field [includes `"browser"`](https://vitejs.dev/config/shared-options.html#resolve-mainfields), but this plugin incorrectly overrides the default with a value which does [not include it](https://github.com/sveltejs/vite-plugin-svelte/blob/main/packages/vite-plugin-svelte/src/utils/constants.js#L3).

The result is that upgrading to Vite 5 and v3 of this plugin causes node build conditions to be imported, and errors when those builds contain imports to node builtins.

### Reproduction URL

https://stackblitz.com/edit/vitejs-vite-tvsiq8?file=.vscode%2Fextensions.json

### Reproduction

- See stackblitz repro
- Run `npm run build`

### Logs

```shell
error during build:
RollupError: "TextEncoder" is not exported by "__vite-browser-external", imported by "node_modules/axios/lib/helpers/formDataToStream.js".
    at error (/home/projects/vitejs-vite-tvsiq8/node_modules/rollup/dist/es/shared/parseAst.js:469:30)
    at error (/home/projects/vitejs-vite-tvsiq8/node_modules/rollup/dist/es/shared/node-entry.js:12753:29)
    at traceVariable (/home/projects/vitejs-vite-tvsiq8/node_modules/rollup/dist/es/shared/node-entry.js:13194:29)
    at findVariable (/home/projects/vitejs-vite-tvsiq8/node_modules/rollup/dist/es/shared/node-entry.js:11607:39)
    at bind (/home/projects/vitejs-vite-tvsiq8/node_modules/rollup/dist/es/shared/node-entry.js:7209:40)
    at bind (/home/projects/vitejs-vite-tvsiq8/node_modules/rollup/dist/es/shared/node-entry.js:4650:23)
    at bind (/home/projects/vitejs-vite-tvsiq8/node_modules/rollup/dist/es/shared/node-entry.js:4650:23)
    at bind (/home/projects/vitejs-vite-tvsiq8/node_modules/rollup/dist/es/shared/node-entry.js:4646:32)
    at bind (/home/projects/vitejs-vite-tvsiq8/node_modules/rollup/dist/es/shared/node-entry.js:4646:32)
[...]

PR #809, opened 2023-11-21
fix: add browser to default resolve.mainFields

WoodyWoodsta:
Fixes https://github.com/sveltejs/vite-plugin-svelte/issues/808


bluwy (approved):
Thanks for catching it, I forgot to fix this 🤦

benoitf:
FYI when using this library https://github.com/Cweili/svelte-fa/blob/master/package.json the update to v3.0.1 make the library failed now (v3.0.1) while with v3.0.0 it was working (with vite5)

dominikg:
> FYI when using this library https://github.com/Cweili/svelte-fa/blob/master/package.json the update to v3.0.1 make the library failed now (v3.0.1) while with v3.0.0 it was working (with vite5)

Hi, please file an issue report with a minimal reproduction. the package.json of that library looks very complex and it's possible that this change unmasked a preexisting issue that was never triggered before due to the previous svelte field resolution algorithm

PR #816, opened 2023-11-28
fix: update Vite peer dependency version

benmccann:
Sets 5.0.3 as the minimum because earlier versions are broken with SvelteKit


dominikg:
wouldn't it make more sense to add it to kit? technically this is q breaking change.

PR #831, opened 2023-12-26
chore: update to SvelteKit 2

bluwy:
- close #825 
- close #826 
- close #829

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,7 +25,7 @@
     "@changesets/cli": "^2.27.1",
     "@svitejs/changesets-changelog-github-compact": "^1.1.0",
     "@types/fs-extra": "^11.0.4",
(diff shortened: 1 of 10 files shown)


package.json
```diff
@@ -66,11 +66,11 @@
   },
   "pnpm": {
     "overrides": {
+      "@sveltejs/kit>@sveltejs/vite-plugin-svelte": "^3.0.0",
```
bluwy, line 69: The override below (`"@sveltejs/vite-plugin-svelte": "workspace:^"`) would also override SvelteKit's v-p-s peer dep, causing peer dep mismatch warnings. This override pins to the original version to avoid the warning.

dominikg (approved):

PR #844, opened 2024-01-26
fix: correctly detect script language during compilation

RisaI:
The script regex introduced in c2017e0 matches comment blocks by mistake. This causes files using a header comment tag to always be detected as `js`.  This will produce an error if the script tag imports a module with an explicit `.js` extensions, where physically only a `.ts` file is present.

This PR removes the comment matching part of the two regex instances and introduces a test to verify the script language gets properly detected in the presence of a header comment block.

## Bug reproduction
```svelte
<!-- this file uses typescript -->
<script lang="ts">
  // the following will cause Failed to resolve import "./foo.js" from "src/*.svelte". Does the file exist?
  // if physically the file's name is `foo.ts`
  import { foo } from './foo.js';
</script>
```

### DIFF
--- a/.changeset/rare-buttons-argue.md
+++ b/.changeset/rare-buttons-argue.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/vite-plugin-svelte': patch
+---
+
+fix(compile): correctly determine script lang in files where a comment precedes the script tag
--- a/packages/vite-plugin-svelte/__tests__/compile.spec.js
+++ b/packages/vite-plugin-svelte/__tests__/compile.spec.js
@@ -45,5 +45,33 @@ describe('createCompileSvelte', () => {
 			);
 			expect(output.compiled.js.code).not.toContain('/* @__PURE__ */\n');
 		});
+
+		it('detects script lang', async () => {
+			const code = `<!-- this file uses typescript -->
+			<script lang="ts" generics="T">
+				const x = 1;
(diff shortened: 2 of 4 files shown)


RisaI:
A similar issue might happen here when detecting style blocks:
https://github.com/sveltejs/vite-plugin-svelte/blob/c36c3aea13754c29279924f0c29f5f141e8747fc/packages/vite-plugin-svelte/src/utils/error.js#L148

dominikg:
Hi, thank you for this PR.

Unfortunately I think it needs adjustments.

As far as I understand, the comment part of the regex is there to prevent it catching script blocks inside a comment.

```html
<!--
<script lang="ts"> 
// deactivated with a comment, ignore it
</script>
-->
<script>
// active, parse it
</script>
```

So instead of using the first match, the previous regex should be kept and then matches need to be iterated until the first script match is found.

This isn't ideal but the best we can do for now. Would you be able to make the adjustment?

RisaI:
Hello, sorry, I didn't realize that was the intent when I first read the code. I added a commented script block to the unit test and made the language detection use `matchAll` to detect the first match with a capturing group. I reverted the change in `error.js`, because the presence of the error is checked by the character indices spanned by the matched string, so no false positives should occur.

dominikg:
thanks a lot! I added a changeset and this will be released shortly.

dominikg (approved):

PR #879, opened 2024-04-01
feat(optimizeDeps): generate globs for svelte 

dominikg:
to increase compatibility with future versions

alternative to #877


dominikg:
@bluwy  mentioned that the globs ending with .js don't match the exports at all, so they are not working as needed. closing

PR #892, opened 2024-04-29
breaking: drop support for svelte4, remove unused code and update types

dominikg:
vite-plugin-svelte v3 supports both svelte4 and svelte5. 

Due to breaking changes in the svelte compiler, around hmr and for how we add svelte-inspector, it is not feasible to support both svelte majors from the same v-p-s major.

Users on svelte4 must continue to use vite-plugin-svelte@3

This PR makes the grunt work of changes to remove svelte4 related code. Expect some tests to fail intially


dominikg:
note to self: check support for lightningcss and css nesting syntax

Conduitry:
The linting errors about `$foo is an illegal variable name` appear to be a bug in the ESLint plugin.

As I've mentioned [here](https://github.com/sveltejs/eslint-plugin-svelte/issues/652), it appears these can be worked around by adding additional references to the stores, like so:

```diff
diff --git a/packages/playground/kit-demo-app/src/routes/Header.svelte b/packages/playground/kit-demo-app/src/routes/Header.svelte
index 368b721e..154eecbb 100644
--- a/packages/playground/kit-demo-app/src/routes/Header.svelte
+++ b/packages/playground/kit-demo-app/src/routes/Header.svelte
@@ -2,6 +2,7 @@
 	import { page } from '$app/stores';
 	import logo from '$lib/images/svelte-logo.svg';
 	import github from '$lib/images/github.svg';
+	page;
 </script>
 
 <header>
diff --git a/packages/playground/kit-demo-app/src/routes/sverdle/+page.svelte b/packages/playground/kit-demo-app/src/routes/sverdle/+page.svelte
index 2784431d..5c07ea55 100644
--- a/packages/playground/kit-demo-app/src/routes/sverdle/+page.svelte
+++ b/packages/playground/kit-demo-app/src/routes/sverdle/+page.svelte
@@ -4,6 +4,8 @@
 
 	import { reduced_motion } from './reduced-motion';
 
+	reduced_motion;
+
 	/** @type {import('./$types').PageData} */
 	export let data;
 
diff --git a/packages/playground/optimizedeps-include/src/App.svelte b/packages/playground/optimizedeps-include/src/App.svelte
index ccbba8da..4564f51d 100644
--- a/packages/playground/optimizedeps-include/src/App.svelte
+++ b/packages/playground/optimizedeps-include/src/App.svelte
@@ -1,5 +1,7 @@
 <script>
 	import { Route, router } from 'tinro';
+
+	router;
 </script>
 
 <nav>
```

This does, then, leave me with a bunch of other type errors still present locally.

I don't know whether it makes sense to apply a workaround like the above for now. I don't know what the expected timeline is on dropping Svelte 4 support here and on fixing the eslint-plugin-svelte bug.

dominikg:
as for the timeline, right before/around svelte5 final release, the eslint bug needs fixing too, users would rightfully be royally annoyed by this i

packages/vite-plugin-svelte/src/utils/load-raw.js
```diff
@@ -19,33 +19,24 @@ export async function loadRaw(svelteRequest, compileSvelte, options) {
 	try {
 		//avoid compileSvelte doing extra ssr stuff unless requested
 		//@ts-ignore //@ts-expect-error generate value differs between svelte4 and 5
```
bluwy, line 21: Looks like we can also remove this

packages/vite-plugin-svelte/package.json
```diff
@@ -41,23 +41,22 @@
   },
   "homepage": "https://github.com/sveltejs/vite-plugin-svelte#readme",
   "dependencies": {
-    "@sveltejs/vite-plugin-svelte-inspector": "^2.1.0",
+    "@sveltejs/vite-plugin-svelte-inspector": "workspace:^",
```
bluwy, line 44: I guess we want to use the static `^3.0.0` after it's published?

packages/vite-plugin-svelte-inspector/package.json
```diff
@@ -44,13 +44,13 @@
     "debug": "^4.3.4"
   },
   "peerDependencies": {
-    "@sveltejs/vite-plugin-svelte": "^3.0.0",
-    "svelte": "^4.0.0 || ^5.0.0-next.0",
+    "@sveltejs/vite-plugin-svelte": "workspace:^",
```
bluwy, line 47: And here (just to highlight)

bluwy (approved):

packages/vite-plugin-svelte/package.json
dominikg: yes, either that, or inline the inspector into v-p-s entirely, or remove the dependency and let users add it to their vite config manually. The current situation is a bit tedious for maintenance

PR #906, opened 2024-05-14
fix: ensure svelte modules correctly run in DEV mode

trueadm:
We were missing this option, meaning Svelte modules that were `.ts` or `.js` were always compiled in prod mode.

### DIFF
--- a/.changeset/long-suns-happen.md
+++ b/.changeset/long-suns-happen.md
@@ -1,0 +1,5 @@
+---
+"@sveltejs/vite-plugin-svelte": major
+---
+
+fix: ensure svelte modules correctly run in DEV mode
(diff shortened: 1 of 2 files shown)


.changeset/long-suns-happen.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sveltejs/vite-plugin-svelte': major
```
dummdidumm, line 2: Is major correct? Yes because repo is in pre release mode currently? Just want to flag to make sure it's not by accident

bluwy: Yeah I think marking as `patch` should be enough since it's unlikely to break users with this change. It'll help with the organization of the generated changelog.

dominikg (approved):

Issue #909, opened 2024-05-15
Error with: `__vite_ssr_import_meta__.hot.acceptExports` when running `vitest` tests through `@testing-library/svelte`


Hugos68:
### Describe the bug

I honestly don't have much knowledge about vite's internals or even public API. I just know that when using `@sveltejs/vite-plugin-svelte@4.0.0-next.1` my tests started breaking.

### Reproduction URL

https://github.com/Hugos68/vite-plugin-svelte-accepts-exports-repro

### Reproduction

See README of the reproduction repository

### Logs

```shell
> vite-plugin-svelte-accepts-exports-repro@0.0.1 test
> vitest run


 RUN  v1.6.0 C:/Users/hugok/Documents/Personal/vite-plugin-svelte-accepts-exports-repro

 ❯ src/index.test.js (0)

[...]

PR #913, opened 2024-05-21
fix: disable hmr when explicitly disabled in vite config

dominikg:
improvement coming from discussion in #909


dominikg:
one caveat is for testsuites that actually do test hmr behavior with vitest, you have to explicitly set compilerOptions.hmr = true.   

these suites are almost guaranteed to be our own or those of other tooling devs very close to us. Users would almost never make custom hmr things that need testing this way.

dominikg:
caveat mentioned previously no longer applies. vitest sets server.hmr to false so we don't have to do anything.

PR #917, opened 2024-05-25
chore: backport fixes

dominikg:
these 2 patches are good to have in v3 too

### DIFF
--- a/packages/vite-plugin-svelte/src/utils/options.js
+++ b/packages/vite-plugin-svelte/src/utils/options.js
@@ -200,10 +200,7 @@ export function resolveOptions(preResolveOptions, viteConfig, cache) {
 		}
 	};
 	const hot =
-		!viteConfig.isProduction &&
-		!preResolveOptions.isBuild &&
-		viteConfig.server &&
-		viteConfig.server.hmr !== false;
+		!viteConfig.isProduction && !preResolveOptions.isBuild && viteConfig.server.hmr !== false;
 	if (isSvelte5) {
 		if (isSvelte5WithHMRSupport) {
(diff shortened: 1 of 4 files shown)


dominikg:
had to refactor the hmr one slightly due to the changes in v4 in the options, but spirit is identical

packages/vite-plugin-svelte/src/utils/options.js
```diff
@@ -199,13 +199,18 @@ export function resolveOptions(preResolveOptions, viteConfig, cache) {
 			dev: !viteConfig.isProduction
 		}
 	};
+	const hot =
+		!viteConfig.isProduction &&
+		!preResolveOptions.isBuild &&
+		viteConfig.server &&
```
bluwy, line 205: Something I just noticed but not a big deal to change. It looks like `viteConfig.server` is always truthy, but this line implies otherwise, which could be confusing because we don't want to disable HMR if the server is not configured (HMR would've been enabled by default in dev)

packages/vite-plugin-svelte/src/utils/options.js
```diff
@@ -235,8 +240,15 @@ export function resolveOptions(preResolveOptions, viteConfig, cache) {
 
 /**
  * @param {import('../types/options.d.ts').ResolvedOptions} options
+ * @param {import('vite').ResolvedConfig} viteConfig
  */
-function enforceOptionsForHmr(options) {
+function enforceOptionsForHmr(options, viteConfig) {
+	if (options.hot && viteConfig.server?.hmr === false) {
```
bluwy, line 246: The `?.` seem unnecessary too.

bluwy (approved):

PR #920, opened 2024-05-27
fix: only apply infix filter to basename

### LINKED ISSUE #919: Infix logic should only apply to basename
### Describe the bug

The infix logic for allowing `*.svelte.*` files (e.g. `thing.svelte.test.js`) to be processed by Svelte fails when the directory contains the infix, for example because the project name is `learn.svelte.dev`. It should only apply to the basename

### Reproduction URL

N/A (will PR, one sec)

### Reproduction

_No response_

### Logs

_No response_

### System Info

```shell
next
```

Rich-Harris:
fixes #919

### DIFF
--- a/packages/vite-plugin-svelte/src/utils/id.js
+++ b/packages/vite-plugin-svelte/src/utils/id.js
@@ -1,5 +1,6 @@
 import { createFilter, normalizePath } from 'vite';
 import * as fs from 'node:fs';
+import * as path from 'node:path';
 import { log } from './log.js';
 import { DEFAULT_SVELTE_MODULE_EXT, DEFAULT_SVELTE_MODULE_INFIX } from './constants.js';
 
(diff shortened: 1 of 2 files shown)


packages/vite-plugin-svelte/src/utils/id.js
```diff
@@ -179,10 +180,15 @@ function buildFilter(include, exclude, extensions) {
...
-		rollupFilter(filename) &&
-		infixes.some((infix) => filename.includes(infix)) &&
-		extensions.some((ext) => filename.endsWith(ext));
+	return (filename) => {
+		const basename = path.basename(filename);
+
+		return (
+			rollupFilter(filename) &&
+			infixes.some((infix) => basename.includes(infix)) &&
+			extensions.some((ext) => filename.endsWith(ext))
```
dummdidumm, line 189: Isn't it better to also use the basename here? Where would that return a false negative?
```suggestion
			extensions.some((ext) => basename.endsWith(ext))
```

Rich-Harris: Identical result, which is why I didn't bother changing it. But certainly no harm in it

bluwy (approved):

PR #924, opened 2024-06-02
chore: update pnpm9

dominikg:
disable package-manager-strict, only allow postinstall for esbuild

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,9 +59,9 @@
       "prettier --cache --ignore-path .gitignore --ignore-unknown --write"
     ]
   },
-  "packageManager": "pnpm@8.15.8",
+  "packageManager": "pnpm@9.1.4+sha512.9df9cf27c91715646c7d675d1c9c8e41f6fce88246f1318c1aa6a1ed1aeb3c4f032fcdf4ba63cc69c4fe6d634279176b5358727d8f2cc1e65b65f43ce2f8bfb0",
   "engines": {
-    "pnpm": "^8.6.3",
+    "pnpm": "^9.0.0",
     "node": "^18.0.0 || >=20"
   },
   "pnpm": {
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -72,6 +72,9 @@
       "svelte": "$svelte",
       "vite": "$vite",
       "@types/node@<=20.12.0": "20.11.5"
-    }
+    },
+    "onlyBuiltDependencies": [
+      "esbuild"
```
dominikg, line 77: pnpm@9 changed the default for enable-pre-post scripts to true. this changes it so that only scripts we trust get run. for now only esbuild.

package.json
```diff
@@ -59,9 +59,9 @@
       "prettier --cache --ignore-path .gitignore --ignore-unknown --write"
     ]
   },
-  "packageManager": "pnpm@8.15.8",
+  "packageManager": "pnpm@9.1.4+sha512.9df9cf27c91715646c7d675d1c9c8e41f6fce88246f1318c1aa6a1ed1aeb3c4f032fcdf4ba63cc69c4fe6d634279176b5358727d8f2cc1e65b65f43ce2f8bfb0",
```
dominikg, line 62: the hash is generated by `corepack up` , you can look it up manually with 

```
curl `npm info pnpm@9.1.4 dist.tarball` | sha512sum
```

it's a bit unfortunate that corepack went that way of attaching a checksum of the tgz, but better than nothing.
I don't use corepack but for those who do, they recommend having the hash for checking purposes.

bluwy (approved):
Was also wondering if CI works with the hash in `packageManager`, and seems like it does: https://github.com/sveltejs/vite-plugin-svelte/actions/runs/9341234007/job/25707770323?pr=924#step:4:9

PR #927, opened 2024-06-11
chore(deps): update to eslint v9

dominikg:
used the migration tool but had to do a few changes after

The `node: false` in globals is replaced by a mapped 'off' spread, but that came after the browser spread, disabling shared globals in the process.

todo: remove the FlatCompat helper


bluwy (approved):
I don't understand every config, but seems ok 😄

dominikg:
@benmccann  does this look ok? esp. the integration of the svelte config? I am a bit surprised that it doesn't setup the ts parser, or does that work ootb now?

dominikg:
oh, and it having browser and node globals enabled for everything by default is a bit lax too o_O

eslint.config.js
```diff
@@ -0,0 +1,259 @@
...
+				'error',
+				{
+					allowModules: ['types', 'estree', 'testUtils', '@sveltejs/vite-plugin-svelte', 'svelte']
+				}
+			],
+
+			'n/no-missing-require': [
+				'error',
+				{
+					allowModules: ['pnpapi']
```
benmccann, line 58: is this still needed? I didn't see `pnpapi` in the code

eslint.config.js
```diff
@@ -0,0 +1,259 @@
...
+			],
+
+			'n/no-extraneous-require': [
+				'error',
+				{
+					allowModules: ['vite']
+				}
+			],
+
+			'n/no-deprecated-api': 'off',
```
benmccann, line 76: are we using many deprecated APIs? maybe we could just ignore them where we are or whitelist them here if they can't be fixed?

eslint.config.js
```diff
@@ -0,0 +1,259 @@
...
+	{
+		files: ['packages/playground/kit-demo-app/src/**'],
+
+		rules: {
+			'n/no-missing-import': 'off',
+			'n/no-unsupported-features/node-builtins': 'off'
+		}
+	},
+	{
+		files: ['**/*.d.ts'],
```
benmccann, line 239: this was already present above

benmccann:
> oh, and it having browser and node globals enabled for everything by default is a bit lax too o_O

I don't remember how many places we'd need to enable them if they're not enabled globally or how hard that would be too do. We could investigate. As long as we don't make it so annoying that Rich rips out the whole eslint setup out of frustration :laughing:

eslint.config.js
dominikg: honestly don't remember why we had this. @bluwy  was it sth that moved with vitefu?

bluwy: Seems to exist for a very long time when the plugin is copied from the Vite repo. Likely safe to remove this now.

eslint.config.js
```diff
@@ -0,0 +1,167 @@
+import markdown from 'eslint-plugin-markdown';
+import globals from 'globals';
+import n from 'eslint-plugin-n';
+import svelte_config from '@sveltejs/eslint-config';
+import svelteParser from 'svelte-eslint-parser';
```
benmccann, line 5: this is camelCase but line above is snake_case

PR #936, opened 2024-06-24
chore: replace execa with cross-spawn

dominikg:
cross-spawn is the dep of execa that does most of what we need anyways, so cut the middleman.


dominikg:
staying on execa v8 until a smaller alternative comes along

PR #941, opened 2024-07-08
feat(config): dynamically generate list of svelte exports

dominikg:
port of #877 for svelte5 only

pulling the list of export names from package.json helps when svelte adds a new export. Otherwise we have to release a new version of vite plugin svelte each time


dominikg:
added a small optimization to exclude server-only exports from optimizeDeps

PR #949, opened 2024-07-25
fix(hmr): compile files in /node_modules/ with hmr: false

dominikg:
this saves some time and space, but also ensures consistency in cssHash which avoids bugs like https://github.com/sveltejs/svelte/issues/12601

### DIFF
--- a/packages/vite-plugin-svelte/src/utils/compile.js
+++ b/packages/vite-plugin-svelte/src/utils/compile.js
@@ -59,6 +59,7 @@ export function createCompileSvelte() {
 		/** @type {import('svelte/compiler').CompileOptions} */
 		const compileOptions = {
 			...options.compilerOptions,
+			hmr: options.compilerOptions.hmr && !normalizedFilename.includes('/node_modules/'),
 			filename,
 			generate: ssr ? 'server' : 'client'
 		};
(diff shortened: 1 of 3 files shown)


Conduitry:
Hm. Could someone conceivably have an `npm link`ed dependency here that they still want HMR for?

trueadm (approved):

packages/vite-plugin-svelte/src/utils/compile.js
```diff
@@ -59,6 +59,7 @@ export function createCompileSvelte() {
 		/** @type {import('svelte/compiler').CompileOptions} */
 		const compileOptions = {
 			...options.compilerOptions,
+			hmr: options.compilerOptions.hmr && !normalizedFilename.includes('/node_modules/'),
```
Conduitry, line 62: TS doesn't like this. It sees that `options.compilerOptions` could potentially be undefined.

dominikg: close enough to false :dagger:  thanks!

Conduitry: No, that would be an actual issue, I think. If `options.compilerOptions` is undefined, then `options.compilerOptions.hmr` is a runtime error.

dominikg:
> Hm. Could someone conceivably have an `npm link`ed dependency here that they still want HMR for?

for that they would have to add the dep to optimizeDeps.exclude or it would be prebundled which prevents hmr too. but for comparability i'll post a different fix that updates cssHash instead.

dominikg:
losing in favor of #950

PR #950, opened 2024-07-26
fix: align css hash in prebundling and disable hmr

dominikg:
alternative to #949  

this preserves hmr for dependencies that are in node_modules but excluded from optimizeDeps and also ensures that cssHash function is applied consistently so that it does not break during hydration

### DIFF
--- a/packages/vite-plugin-svelte/src/utils/id.js
+++ b/packages/vite-plugin-svelte/src/utils/id.js
@@ -133,19 +133,8 @@ function parseRequestQuery(rawQuery) {
  * @param {string} normalizedRoot
  * @returns {string}
  */
-function normalize(filename, normalizedRoot) {
+export function normalize(filename, normalizedRoot) {
 	return stripRoot(normalizePath(filename), normalizedRoot);
-}
-
-/**
- * create a normalizer function for root
- *
- * @param {string} root
- * @returns {(filename: string) => string}
- */
-export function buildFilenameNormalizer(root) {
-	const normalizedRoot = normalizePath(root);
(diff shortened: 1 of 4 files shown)


dominikg:
cc @trueadm  @Conduitry   

this fixes it too but does not prevent hmr in node_modules completely

packages/vite-plugin-svelte/src/utils/id.js
```diff
@@ -137,6 +137,17 @@ function normalize(filename, normalizedRoot) {
...
 }
 
+/**
+ * create a normalizer function for root
+ *
+ * @param {string} root
+ * @returns {(filename: string) => string}
+ */
+export function buildFilenameNormalizer(root) {
+	const normalizedRoot = normalizePath(root);
```
bluwy, line 147: I think we can skip the normalizePath here since Vite already normalizes it. We pass the root from https://github.com/sveltejs/vite-plugin-svelte/blob/bd915c1317127a716a22cdd8c44d66d766e0f513/packages/vite-plugin-svelte/src/utils/options.js#L212

bluwy (approved):

dominikg:
hmr inside node_modules is blocked by 2 other problems, one being that node_modules is on vites watch ignore list by default, the other that at least with pnpm, handleHotUpdate gets called with symlink name and empty modules list so that the update is discarded by vite-plugin-svelte 

see https://github.com/vitejs/vite/issues/8619#issuecomment-2254154530

we can still go ahead with this PR as it is a quite minimal fix for the problem described by https://github.com/sveltejs/svelte/issues/12601

PR #959, opened 2024-08-09
chore: publish package provenance info

benmccann:
I tested this on SvelteKit already. Now we have a nice little green checkmark next to the version on npm

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -10,6 +10,7 @@ jobs:
   release:
     permissions:
       contents: write # to create release (changesets/action)
+      id-token: write # OpenID Connect token needed for provenance
       pull-requests: write # to create pull request (changesets/action)
     # prevents this action from running on forks
     if: github.repository == 'sveltejs/vite-plugin-svelte'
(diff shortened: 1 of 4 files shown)


packages/vite-plugin-svelte-inspector/package.json
```diff
@@ -52,5 +52,8 @@
     "@types/debug": "^4.1.12",
     "svelte": "^5.0.0-next.210",
     "vite": "^5.3.5"
+  },
+  "publishConfig": {
+    "provenance": true
```
dominikg, line 57: is there a reason to add this to multiple package.json files instead of setting it in the workspace .npmrc? 

https://docs.npmjs.com/generating-provenance-statements#using-third-party-package-publishing-tools

.github/workflows/release.yml
```diff
@@ -10,6 +10,7 @@ jobs:
   release:
     permissions:
       contents: write # to create release (changesets/action)
+      id-token: write # OpenID Connect token needed for provenance
```
dominikg, line 13: what powers exactly does this token grant. is it possible to restrict it only to those required for provenance to work?

see https://docs.github.com/en/actions/security-for-github-actions/security-hardening-your-deployments/about-security-hardening-with-openid-connect#customizing-the-token-claims

cc @pngwn , is this something that needs to be protected from extraction too?

dominikg (approved):

PR #962, opened 2024-08-18
fix(peerDeps): remove experimental svelte5 support from v3

dominikg:
to avoid peer dependency mismatch with svelte-hmr

Svelte5 users can upgrade to next major `"@sveltejs/vite-plugin-svelte": "^4.0.0-next.6"` in their devDependencies


dominikg:
closing in favor of #964

PR #966, opened 2024-08-22
chore(actions): harden release action and clean up ci

dominikg:
- `--prefer-offline` isn't needed
-  pinning changesets/action prevents malicious updates creeping in
- remove `.npmrc` after use to avoid it leaking token


pngwn (approved):
Looks great! Nice catch about the npmrc.

PR #983, opened 2024-09-15
fix(deps): added missing spacing around ||

skdev-io:
Issue: https://github.com/sveltejs/vite-plugin-svelte/issues/984

@sveltejs/vite-plugin-svelte-inspector version was missing spacing around "||".

### DIFF
--- a/packages/vite-plugin-svelte/package.json
+++ b/packages/vite-plugin-svelte/package.json
@@ -41,7 +41,7 @@
   },
   "homepage": "https://github.com/sveltejs/vite-plugin-svelte#readme",
   "dependencies": {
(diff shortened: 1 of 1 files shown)


Kapsonfire-DE:
I checked the docs and dont see the space is mandatory in package.json - therefore it would be a bun bug and in responsibility of the bun team to fix this.

Kapsonfire-DE:
> [dependencies](https://docs.npmjs.com/cli/v10/configuring-npm/package-json#dependencies)
Dependencies are specified in a simple object that maps a package name to a version range. The version range is a string which has one or more space-separated descriptors. Dependencies can also be identified with a tarball or git URL.


good catch!

dominikg:
bun fixed it and we are removing the || separator in the propper 3.0 release. Thank you for bringing it to attention!!

PR #995, opened 2024-10-17
breaking(options): change default key-combo for svelte-inspector to alt-x

dominikg:
Why `alt-x`?

It should be 2 key, and usable with one hand. And most importantly not conflict with a lot of other key-combos.

The previous defaults (ctrl-shift or meta-shift) are used in many combos so they accidentally triggered the inspector eg when making a screenshot with meta-shift-3 on mac.

During my testing, `alt-x` did not trigger anything in linux, windows, mac for chrome or firefox, which should cover most users setups for development.

### DIFF
--- a/packages/vite-plugin-svelte-inspector/src/runtime/Inspector.svelte
+++ b/packages/vite-plugin-svelte-inspector/src/runtime/Inspector.svelte
@@ -4,8 +4,10 @@
 
 	import options from 'virtual:svelte-inspector-options';
 	const toggle_combo = options.toggleKeyCombo?.toLowerCase().split('-');
-	const escape_keys = options.escapeKeys?.map((key) => key?.toLowerCase());
-	const nav_keys = Object.values(options.navKeys).map((k) => k.toLowerCase());
+	const escape_keys = options.escapeKeys?.map((k) => k.toLowerCase());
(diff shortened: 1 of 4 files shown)


packages/vite-plugin-svelte-inspector/src/runtime/Inspector.svelte
```diff
@@ -134,67 +136,67 @@
...
 		switch (key) {
 			case 'shift':
 			case 'control':
 			case 'alt':
 			case 'meta':
-				return event.getModifierState(key.charAt(0).toUpperCase() + key.slice(1));
+				return e.getModifierState(key.charAt(0).toUpperCase() + key.slice(1));
 			default:
-				return key === event.key.toLowerCase();
+				return key === e.code.replace(/^Key/, '').toLowerCase() || key === e.key.toLowerCase();
```
dominikg, line 147: this change helps detecting the physical key if the modifier changed what event.key is. (eg alt-x produces a double-tilde on mac)

took the liberty to slightly refactor the is_ helpers below to all use this logic to avoid inconsistencies.
Using toLowerCase allows the user to use mixed casing in their config, `Alt-X`, `alt-x` and `alt-X` all work the same

PR #1010, opened 2024-10-28
fix: remove references to compiler options no longer available in svelte5

### LINKED ISSUE #1002: Remove references to `hydratable` compiler option in docs
### Describe the problem

Svelte 5 had removed the option as components are always hydratable by default now. However the docs, and a couple places in code, still referenced it.

Remove references to `hydratable` compiler option in docs (and some code)

### Alternatives considered

Not breaking anything and can be ignored, but could be misleading.

### Importance

nice to have

dominikg:
hydratable and enableSourcemaps were removed, also values for generate and css have changed.

fixes #1002

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -240,9 +240,9 @@ A [picomatch pattern](https://github.com/micromatch/picomatch), or array of patt
     plugins: [
       svelte({
         dynamicCompileOptions({ filename, compileOptions }) {
-          // Dynamically set immutable flag per Svelte file
-          if (compileWithImmutable(filename) && !compileOptions.immutable) {
-            return { immutable: true };
+          // Dynamically set runes mode per Svelte file
+          if (forceRunesMode(filename) && !compileOptions.runes) {
+            return { runes: true };
           }
         }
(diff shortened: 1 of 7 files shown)


docs/config.md
```diff
@@ -240,9 +240,9 @@ A [picomatch pattern](https://github.com/micromatch/picomatch), or array of patt
     plugins: [
       svelte({
         dynamicCompileOptions({ filename, compileOptions }) {
-          // Dynamically set hydration per Svelte file
-          if (compileWithHydratable(filename) && !compileOptions.hydratable) {
-            return { hydratable: true };
+          // Dynamically set immutable flag per Svelte file
+          if (compileWithImmutable(filename) && !compileOptions.immutable) {
+            return { immutable: true };
```
bluwy, line 245: It looks like immutable is also deprecated when using svelte with runes. Do we also want to avoid referring to it? https://svelte.dev/docs/svelte/compiler-warnings#optionsdeprecatedimmutable

bluwy (approved):

PR #1020, opened 2024-11-08
feat!: vite 6 support

bluwy:
Very rough support. There's some things I want to double-check first so marking as draft. But this should pass vite-ecosystem-ci now.


packages/e2e-tests/vite-ssr-esm/vite.config.js
```diff
@@ -14,6 +14,11 @@ export default defineConfig(({ command, mode }) => {
 				}
 			}
 		},
+		// TODO: investigate the condition issue. it's the same thing dominik and ben found.
+		// idk why it's only happening after https://github.com/vitejs/vite/pull/18395
+		ssr: {
+			noExternal: ['esm-env']
```
dominikg, line 20: if this is a general issue, should vite-plugin-svelte add that to the users config by default?!

packages/vite-plugin-svelte/src/index.js
```diff
@@ -67,14 +63,31 @@ export function svelte(inlineOptions) {
...
+				}
+				config.resolve.mainFields.unshift('svelte');
+
+				// Emulate Vite default fallback for `resolve.conditions` if not set
 				if (config.resolve.conditions == null) {
-					if (name === 'client') {
-						config.resolve.conditions = [...VITE_CLIENT_RESOLVE_CONDITIONS];
+					// These exports only exist in Vite 6
+					const { defaultClientConditions, defaultServerConditions } = await import('vite');
+					if (name === 'client' || opts.isSsrTargetWebworker) {
```
dominikg, line 87: is name === 'client' a good way to check this? that would assume there can only be one client environment in an app.

dominikg: what if there was a tauri app and a web app at the same time, or two different clients?

packages/vite-plugin-svelte/src/index.js
```diff
@@ -67,14 +63,31 @@ export function svelte(inlineOptions) {
...
-			configEnvironment(name, config) {
+
+			// @ts-ignore This hook only works in Vite 6
+			async configEnvironment(name, config, opts) {
 				config.resolve ??= {};
+
+				// Emulate Vite default fallback for `resolve.mainFields` if not set
+				if (config.resolve.mainFields == null) {
+					// These exports only exist in Vite 6
+					const { defaultClientMainFields, defaultServerMainFields } = await import('vite');
```
dominikg, line 74: this or `import * as vite from 'vite'` ?

packages/vite-plugin-svelte/src/utils/options.js
```diff
@@ -334,22 +334,24 @@ function resolveViteRoot(viteConfig) {
...
 
 	/** @type {Partial<import('vite').UserConfig>} */
 	const extraViteConfig = {
 		resolve: {
 			dedupe: [...SVELTE_IMPORTS],
-			// In Vite 6, we need to provide the default conditions too as it now replaces the default,
-			// instead of extending it. We set undefined here and extend it in the `configEnvironment` hook instead.
+			// In Vite 6, conditions now override the defaults instead of extending. `resolve.conditions`
+			// also only apply to the client env, so we use the `configEnvironment` hook to set it up
+			// instead, so here we set to `undefined` to skip it.
```
dominikg, line 354: i am still not sure i follow why the global `resolve.conditions` is not extended into all environments by default.

packages/vite-plugin-svelte/src/index.js
bluwy: Ah true, I think it should check the `consumer` here, which is the same as Vite internally.

packages/vite-plugin-svelte/src/index.js
bluwy: I'd lean on this for now I think so when/if we only support Vite 6, we get cleaner named imports

packages/e2e-tests/vite-ssr-esm/vite.config.js
dominikg: thats cheating the ecosystem-ci, users apps will run into this very issue then....

dominikg: so we definetely need to figure this out before supporting vite 6

packages/vite-plugin-svelte/src/utils/options.js
bluwy: It now only applies to the client environment. Most of the top-level options now focuses on the client environment only now, but other environments could reuse it, e.g. `build.sourcemap`, but not `build.modulePreload`.

I think the decision was to make it not ambiguously apply to client and ssr environments simultaneously so it's easier to reason about how the options get inherited and applied. But if you ask me, I'm not sure this is a good behaviour either, but if we want to fix this, we'd likely have to change a lot more of the environment API to make the reasoning easier again, and that's a lot of work for me to bring research and suggest a solution again 😅

packages/vite-plugin-svelte/src/index.js
bluwy: I'm not sure if it's needed imo

packages/e2e-tests/vite-ssr-esm/vite.config.js
dominikg: it could get lost in the storm, prefer to let this fail until we have a fix in vps. given svelte itself uses esm-env and sveltekit iirc also did the noExternal thing for it that should be an easy one. But i am also not convinced this fixes it in all projects unless users add the right conditions to their starting arguments.

bluwy: Removed 👍 Yeah given SvelteKit also did that, maybe we could also do that for now and call it a day. But I think the source of the problem still boils down to `esm-env` not working well with Vite without `--conditions` properly set, so we should look at this as not only a v-p-s problem.

packages/vite-plugin-svelte/src/index.js
dominikg: well how are others going to use these exposed constants then? you'll always want the ones applicable to the environment you are in, no?

packages/vite-plugin-svelte/src/index.js
dominikg: as discussed on discord we could also just bump vite to 6 and release a major

packages/vite-plugin-svelte/src/utils/options.js
dominikg: this is a pretty hefty change that will break backwards compat of configs for fallback environments?

packages/vite-plugin-svelte/src/index.js
bluwy: I'm not really sure what you mean. Do you have a pseudocode in mind for this?

packages/vite-plugin-svelte/src/index.js
dominikg: i'd really prefer a new major of vps for this, otherwise we have to extend the test matrix to cover vite5 and 6 too

bluwy: So should we start the prerelease on `main` now or a different branch?

dominikg: update deps on main first and release the pending release even though its mostly cosmetic so that 4.0.1 goes out from main. after that we have to branch of v4 like we did for the previous ones and then start pre on main

bluwy: `main` is now in pre mode

packages/vite-plugin-svelte/src/index.js
bluwy: Looks like I need a combination of `config.consumer === 'client'` and `name === 'client'` checks to work, since `config.consumer` could be undefined but it's documented to fallback to `consumer = 'client'` if `name` is `'client'`

packages/e2e-tests/kit-node/__tests__/kit.spec.ts
```diff
@@ -285,7 +285,8 @@ describe('kit-node', () => {
 						'/^svelte\\//', // serialized with toString
 						'e2e-test-dep-svelte-api-only',
 						'svelte-i18n',
-						'esm-env',
+						'esm-env', // first added by svelte-kit
+						'esm-env', // second added by vite-plugin-svelte
```
dominikg, line 289: can we add a check to avoid the duplication or does it not cause any trouble w vite?

packages/vite-plugin-svelte/src/utils/options.js
```diff
@@ -579,6 +572,11 @@ function buildExtraConfigForSvelte(config) {
 	if (!isDepExternaled('svelte', config.ssr?.external ?? [])) {
 		noExternal.push('svelte', /^svelte\//);
 	}
+	// esm-env needs to be bundled by default for the development/production condition
+	// be properly used by svelte
+	if (!isDepExternaled('esm-env', config.ssr?.external ?? [])) {
+		noExternal.push('esm-env');
+	}
```
dominikg, line 579: add warning log that externalizing it will cause trouble?

(with filterable code)

dominikg (approved):

packages/e2e-tests/kit-node/__tests__/kit.spec.ts
```diff
@@ -351,7 +352,7 @@ describe('kit-node', () => {
 				);
 				expectArrayEqual(
 					config.resolve.conditions,
-					['svelte'],
+					['browser', 'development|production', 'module', 'svelte'],
```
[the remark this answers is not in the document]
bluwy: Yeah we should use the vite export here, will update

packages/vite-plugin-svelte/src/utils/options.js
bluwy: I'd personally defer this to later if someone stumbles on the issue 🤔

dominikg: ok, lets hope noone externalizes it.

PR #1024, opened 2024-11-15
feat: support `?inline` query on svelte style virtual modules

bluwy:
In Vite 6, importing CSS in SSR now no longer return a default export of the transformed CSS file, you'd need to use `?inline` instead. This was already pushed for in Vite 5 but we forgot to remove support in SSR.

That means frameworks who do CSS crawling to prevent FOUC will now have to use `?inline` to get the transformed CSS, which most already did. However, there's a quirk where v-p-s always strips out the `?inline` here: https://github.com/sveltejs/vite-plugin-svelte/blob/da54670065f949c7352597dcafbd51dfa202e638/packages/vite-plugin-svelte/src/index.js#L127-L136

This was fine for Vite 5 as mentioned above, Vite still supported default exports of CSS files in SSR, but with Vite 6 it is now empty. For v-p-s to properly support `?inline`, it needs to leave the query as is, similar to `?raw` handling.

---

In practice, this isn't a breaking change, however due to some unintended reliance of the old behaviour, **the SvelteKit tests here will fail due to HMR**. I sent a PR upstream to fix this https://github.com/sveltejs/kit/pull/13007 and if we want to play it safe for now, we can merge this only in the next major.

With this merged, it should unblock Vite 6 support for CSS FOUC for both SvelteKit and Astro.

### DIFF
--- a/packages/e2e-tests/import-queries/__tests__/import-queries.spec.ts
+++ b/packages/e2e-tests/import-queries/__tests__/import-queries.spec.ts
@@ -162,4 +162,11 @@ describe.runIf(!isBuild)('ssrLoadModule', () => {
 		const result = await ssrLoadDummy('?raw&svelte&type=style');
 		await expect(result).toMatchFileSnapshot(snapshotFilename('ssr-style'));
 	});
+	test('?inline&svelte&type=style&lang.css', async () => {
+		// Preload Dummy.svelte first so its CSS is processed in the module graph, otherwise loading
+		// its css inlined url directly will return the raw svelte file rather than the style
+		await ssrLoadDummy('');
+		const result = await ssrLoadDummy('?inline&svelte&type=style&lang.css');
+		await expect(result).toMatchFileSnapshot(snapshotFilename('ssr-inline-style'));
+	});
 });
--- a/packages/vite-plugin-svelte/src/utils/id.js
+++ b/packages/vite-plugin-svelte/src/utils/id.js
@@ -50,7 +50,8 @@ function parseToSvelteRequest(id, filename, rawQuery, root, timestamp, ssr) {
 		query,
 		timestamp,
 		ssr,
-		raw
+		raw,
+		inline: !!query.inline
 	};
 }
 
--- a/.changeset/sixty-chefs-relate.md
+++ b/.changeset/sixty-chefs-relate.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/vite-plugin-svelte': minor
+---
+
+Support `?inline` query on Svelte style virtual modules
--- a/packages/e2e-tests/import-queries/__tests__/__snapshots__/svelte-5/ssr-inline-style.txt
+++ b/packages/e2e-tests/import-queries/__tests__/__snapshots__/svelte-5/ssr-inline-style.txt
@@ -1,0 +1,3 @@
+button.svelte-d8vj6a {
+  color: #000099;
+}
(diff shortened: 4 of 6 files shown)


packages/vite-plugin-svelte/src/utils/id.js
```diff
@@ -50,7 +50,8 @@ function parseToSvelteRequest(id, filename, rawQuery, root, timestamp, ssr) {
 		query,
 		timestamp,
 		ssr,
-		raw
+		raw,
+		inline: !!query.inline
```
dominikg, line 54: raw and direct are handled above, is inline distinct to warrant its own flag? query is returned above too so the consumer could also use query.inline

bluwy: I updated to use `svelteRequest.query.inline` instead. I initially saw `raw` and was trying to map the same pattern here.

packages/e2e-tests/kit-node/vite.config.js
```diff
@@ -15,11 +15,33 @@ const config = {
...
+	],
 	optimizeDeps: {
 		// eagerly include these, otherwise vite optimizer might interfere with restarting while the test is running
 		include: ['svelte-i18n', 'e2e-test-dep-svelte-api-only']
 	}
 };
 
-export default config;
+/**
+ * Workaround until https://github.com/sveltejs/kit/pull/13007 is merged
```
dominikg, line 31: lets merge the kit PR first then, really not a fan of extra code in tests to work around issues in the setup being tested. users don't have the same luxury so this is just hiding an issue. 

would rather merge this with a failed test than hiding it if we can't get the kit PR released in time

packages/e2e-tests/import-queries/__tests__/import-queries.spec.ts
```diff
@@ -162,4 +162,11 @@ describe.runIf(!isBuild)('ssrLoadModule', () => {
 		const result = await ssrLoadDummy('?raw&svelte&type=style');
 		await expect(result).toMatchFileSnapshot(snapshotFilename('ssr-style'));
 	});
+	test('?inline&svelte&type=style&lang.css', async () => {
+		// Preload Dummy.svelte first so its CSS is processed in the module graph, otherwise loading
```
dominikg, line 166: wait, wouldn't this also affect real use if the inline style is explicitly asked for without the .svelte being processed first? The solution here would be to compile the file if cache is empty instead of just bailing.

packages/e2e-tests/kit-node/vite.config.js
bluwy: I don't think it's hiding since there's already a PR out to track this. If it's still not getting reviews I don't think we should be sacrificing our CI for it.

packages/e2e-tests/import-queries/__tests__/import-queries.spec.ts
bluwy: If the virtual styles (inline or not) is requested before the main module is compiled, it'll not work like before. If we want to allow accessing the virtual styles directly and automatically compile the main module, I think that's a separate feature. Was also discussed in https://github.com/sveltejs/vite-plugin-svelte/issues/341

dominikg: makes sense to put it in a separarate PR. :+1:

packages/e2e-tests/kit-node/vite.config.js
dominikg: added an issue to remind us to remove this

dominikg (approved):

PR #1026, opened 2024-11-18
feat: support vite 6 and enter prerelease

bluwy:


bluwy:
Actual Vite dependency not bumped yet as CI will fail until https://github.com/sveltejs/vite-plugin-svelte/pull/1020 is merged.  Only bumped the peer dep ranges

PR #1028, opened 2024-11-21
fix: use script lang regex from svelte compiler

fehnomenal:
Fix for https://github.com/sveltejs/svelte/issues/14384

I just copied the svelte compiler's regex and adjusted the match array index.

Question: is it possible to directly use the regex from the svelte compiler?

### DIFF
--- a/.changeset/swift-dolphins-cheat.md
+++ b/.changeset/swift-dolphins-cheat.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/vite-plugin-svelte': patch
+---
+
+Allow script tags to span multiple lines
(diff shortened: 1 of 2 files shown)


bluwy (approved):

dominikg:
good to go as a quick fix but we really need to do sth about the regex in general

bluwy:
I've merged this PR to `main`, but it looks like GitHub is down or something and not registering it yet, so the PR looks open now. The commit is already in `main`.

bluwy:
I just tried to merge again and it seems to work now, but it created an empty commit in `main`.

PR #1029, opened 2024-11-21
chore: add generate:types command to lint-staged

bluwy:


package.json
```diff
@@ -19,7 +19,8 @@
     "release": "pnpm changeset publish",
     "prepare": "husky",
     "playwright": "playwright-core",
-    "generate:types": "pnpm --filter \"./packages/*\" --parallel generate:types"
+    "generate:types": "pnpm --filter \"./packages/*\" --parallel generate:types",
+    "generate:types-staged": "pnpm generate:types && :"
```
bluwy, line 23: This separate command was needed as lint-staged will passed changed files to the command by default, and the `:` will toss them away. Don't know if Windows will be happy about this though.

I can't inline this directly into lint-staged, it seems to not like the `&&`.

dominikg: hmmm, can you consume the input in the generate script instead?

bluwy: Do you mean putting the `&& :` directly in `generate:types` instead? Yeah that works too if you're ok with that.

dominikg: no, i meant putting it in a node script that handles it, calling dts-buddy from there.

dominikg (approved):

PR #1044, opened 2024-12-11
fix: update warning codes for svelte5

dominikg:
needs to be backported to v-p-s@4 too

### DIFF
--- a/packages/vite-plugin-svelte/src/utils/error.js
+++ b/packages/vite-plugin-svelte/src/utils/error.js
@@ -102,6 +102,16 @@ function formatFrameForVite(frame) {
 }
 
 /**
+ *
+ * @param {string} code the svelte error code
+ * @see https://github.com/sveltejs/svelte/blob/main/packages/svelte/src/compiler/errors.js
+ * @returns {boolean}
+ */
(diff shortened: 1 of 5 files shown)


packages/vite-plugin-svelte/src/utils/error.js
```diff
@@ -114,6 +114,7 @@ export function enhanceCompileError(err, originalCode, preprocessors) {
 
 	// Handle incorrect CSS preprocessor usage
 	if (err.code === 'css-syntax-error') {
```
dominikg, line 116: @dummdidumm  whats the new way to identify css syntax errors? i looked but didn't find the corrosponding code in svelte5

dominikg:
see https://github.com/sveltejs/svelte/issues/14654

dummdidumm (approved):

packages/vite-plugin-svelte/src/utils/error.js
dominikg: added with generic startswith css_ , technically this could introduce false positives for users who don't use a css preprocessor, so if we want to play it conservative, we'd have to use a hardcoded list at the cost of that getting out of date if new codes are added.

PR #1054, opened 2024-12-18
Use optional chaining to catch null error codes in error.js

trbernstein:
Starting a new vite project and selecting svelte-ts template results in an error, because `code` var in error.js is null. Optional chaining resolves this issue.


trbernstein:
Solves #1055.

dominikg (approved):

PR #1058, opened 2024-12-20
fix(types): correct default key combo in typing docs

hyrious:
The default key combo was changed in #995 but not reflect in TypeScript definitions.


bluwy (approved):

bluwy:
Thanks!

PR #1061, opened 2025-01-02
Made overlay a bit nicer

Joedotmt:
Increased padding
changed font to monospace/sans-serif


dominikg:
Hi, thanks for your input. Styles are subjective and the current small padding has been chosen to not cover up too much of the ui below.

If you want to change styles, you can load a separate stylesheet during dev and override the styles for `#svelte-inspector-overlay` like this https://www.sveltelab.dev/inuhu4rvz27axl3

PR #1092, opened 2025-03-02
feat: scope css to js module to allow treeshaking it (requires vite 6.2)

dominikg:
see https://github.com/vitejs/vite/pull/19418

This allows vite to treeshake unused svelte component css if the component was referenced in a barrel file.

### DIFF
--- a/packages/vite-plugin-svelte/src/utils/compile.js
+++ b/packages/vite-plugin-svelte/src/utils/compile.js
@@ -68,19 +68,18 @@ export function createCompileSvelte() {
 		}
 
 		let preprocessed;
-
-		const preprocessors = options.preprocess
-			? Array.isArray(options.preprocess)
-				? [...options.preprocess]
-				: [options.preprocess]
-			: [];
-
+		let preprocessors = options.preprocess;
 		if (!options.isBuild && options.emitCss && compileOptions.hmr) {
 			// inject preprocessor that ensures css hmr works better
-			preprocessors.push(devStylePreprocessor);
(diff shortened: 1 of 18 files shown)


packages/vite-plugin-svelte/src/utils/compile.js
```diff
@@ -133,6 +133,31 @@ export function createCompileSvelte() {
 		let compiled;
 		try {
 			compiled = svelte.compile(finalCode, { ...finalCompileOptions, filename });
+			if (compiled.css) {
+				if (finalCode.includes(':global') && compiled.ast.css) {
```
bluwy, line 137: Could we have a comment here of what this is trying to do? Is it checking for query selectors that starts with `:global()` and not the nested kind?

Also a bit concerned of the perf implication with the additional walking but I guess it can be optimized later.

dominikg: yes, unscoped `:global` must be included or it would be a breaking change and also differ in experience between dev and build because in dev vite still includes the css, its only treeshaken in build.

the include check before walking should minimize impact. If you feel this is still too much of a regression, should we make this opt-in?

bluwy (approved):

dominikg:
the regex approach is a bit less clunky i think. least clunk inside vps would be an option that allows users to define the tests themselves on filename+content, similar to dynamicCompileOptions.

dominikg:
reninder to self: switch to css.hasGlobal https://github.com/sveltejs/svelte/pull/15450#event-17216598471

PR #1095, opened 2025-03-05
chore: update `eslint-plugin-svelte` to v3

baseballyama:


package.json
```diff
@@ -38,7 +38,7 @@
     "eslint-plugin-markdown": "^5.1.0",
     "eslint-plugin-n": "^17.16.2",
     "eslint-plugin-prettier": "^5.2.3",
-    "eslint-plugin-svelte": "^2.46.1",
+    "eslint-plugin-svelte": "^3.0.3",
```
baseballyama, line 41: Most ESLint-related dependencies are likely installed via `@sveltejs/eslint-config`, so I don’t think we need to add them in this repository. If necessary, I will address it in a separate PR.

packages/playground/big/jsconfig.json
baseballyama, line 1: The `big` project wasn’t being checked at all because there was no `jsconfig.json`, so I’ve added it.

packages/playground/big/src/App.svelte
```diff
@@ -13,7 +13,9 @@
 <nav>
 	<h2>This is a huge app</h2>
 	{#each Object.entries(pages) as [name, cmp] (cmp)}
-		<a class="App-link" href on:click|preventDefault={go(cmp)}>{name.replace('_', ' ')}</a>
+		<a class="App-link" href={`?page=${name}`} on:click|preventDefault={go(cmp)}
```
baseballyama, line 16: I'm not sure but is `href` correct? After adding `jsconfig.json`, type errors occurred, so I’ve added a temporary href.

eslint.config.js
```diff
@@ -120,13 +120,6 @@ export default [
 			'prefer-const': 'off' // this turns let foo = $derived into a const otherwise
 		}
 	},
-	{
-		name: 'local/svelte-runes-globals',
-		files: ['**/*.svelte.js', '**/*.svelte.ts', '**/*.svelte.*.js', '**/*.svelte.*.ts'],
-		languageOptions: {
-			parser: svelteParser
-		}
```
baseballyama, line 128: I think this setting is unnecessary because it’s handled by `@sveltejs/eslint-config`.

packages/playground/kit-demo-app/src/routes/sverdle/+page.svelte
```diff
@@ -67,9 +67,11 @@
 		const key = /** @type {HTMLButtonElement} */ (event.target).getAttribute('data-key');
 
 		if (key === 'backspace') {
+			// eslint-disable-next-line svelte/no-reactive-reassign
```
dominikg, line 70: is this present in the current `sv create` template as well? if yes we should fix it there too. But i hope it uses code that doesn't require this and we can just recreate kit-demo-app here.

eslint.config.js
```diff
@@ -15,7 +14,8 @@ export default [
 			'packages/playground/big/src/pages/**', // lots of generated files
 			'packages/*/types/index.d.ts',
 			'packages/*/types/index.d.ts.map',
-			'packages/*/CHANGELOG.md'
+			'packages/*/CHANGELOG.md',
```
dominikg, line 17: i assume these are added to avoid the "embedded svelte" errors coming from  markdown codefence sources? Would be good to have those working

baseballyama: OK. I will adjust it.

eslint.config.js
```diff
@@ -15,7 +14,8 @@ export default [
...
+			'docs/**/*.svelte'
```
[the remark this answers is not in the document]
dominikg: some rules like "missing import" are disabled in markdown, but i'd prefer if linting isn't disabled completely to ensure code in documentation passes lint rules

packages/playground/kit-demo-app/src/routes/sverdle/+page.svelte
baseballyama: The CLI repository uses Svelte5 Runes, allowing for more accurate checks, so ESLint errors shouldn’t occur.

packages/playground/big/src/App.svelte
dominikg: this is an old test app with many generated components made by @rixo  i assume that because on:click calls `go` the href isn't wanted, if you need to use a value it could be`href="#"` or sth inert that doesn't mess with the basic "routing" in the app

baseballyama: `#` shows a compiler warning.

image

eslint.config.js
baseballyama: updated: https://github.com/sveltejs/vite-plugin-svelte/pull/1095/commits/cf5db5c983a8cdd7c4d73b9c3bc075bc6bff07f8

dominikg: great!

packages/playground/kit-demo-app/src/routes/sverdle/+page.svelte
dominikg: we should update kit-demo-app to svelte5 syntax here then too. but that can be done in a separate PR

dominikg (approved):

PR #1099, opened 2025-03-16
Add node exports for CJS

ricardo-devis-agullo:
Since the newest versions of node supports require(esm), adding this exports make that work, (without it, requiring fails with a ERR_PACKAGE_PATH_NOT_EXPORTED). I need for one of my projects that still can't be moved from CJS and AFAIK adding this extra definition should do no harm.

### DIFF
--- a/packages/vite-plugin-svelte/package.json
+++ b/packages/vite-plugin-svelte/package.json
@@ -14,6 +14,10 @@
       "import": {
         "types": "./types/index.d.ts",
         "default": "./src/index.js"
+      },
+      "node": {
+        "types": "./types/index.d.ts",
+        "default": "./src/index.js"
       }
     }
   },


dominikg:
i don't think this is needed. If at all we would have to use the `module-sync` condition https://github.com/nodejs/node/pull/54648

but you can use vite plugin svelte in cjs today if you really have to: https://github.com/sveltejs/vite-plugin-svelte/blob/main/docs/faq.md#how-can-i-use-vite-plugin-svelte-from-commonjs

ricardo-devis-agullo:
Oh, I didn't know about the module-sync condition. Thanks! The problem of the current solution proposed is that it requires me to do an async operation (awaiting the import), and on the constraints that I work with I **need** the import to be synchronous. I understand if from a principle base you wouldn't want to add this to the package.json to disallow CJS imports and if so, I can fork this plugin for my needs, but is there any practical downsides of adding this to the package.json?

Let me know if you want me to edit this PR or we can just close it for now, and thanks for replying so quickly! :)

dominikg:
can you share where you need it sync? vites defineConfig accepts an async function so you can require vite.config.js and it'll still work with dynamic imports inside

ricardo-devis-agullo:
I'll admit it's not a pretty usual case, but I'm not using vite.configs directly. This is for a framework that accepts vite base templates to render as microfrontends. For that it just does internally a "vite build". The template has a strict api that just needs to have an exported function "compile" with an api like this:

module.exports.compile = createCompile({
  plugins: [svelte()],
  // ... more options
});

this file will be automatically required by the framework when pointing the template you are using (will look for node_modules and do it itself). Because of these constraints I can't put anywhere on that file a promise, since it won't resolve.

It's kind of niche, but I thought it wouldn't break anything to explicitly put the exports for cjs. For now I already put a workaround as a postinstall script to add it on the template scaffold.

dominikg:
another option proposed by @bluwy  is to change `"import"` to `"default"`, which would lead to a different error if you tried to require vite-plugin-svelte in a node version that does not support require-esm, but would end up making this work.

I still believe `"import"` to be the technically correct condition to use for vite-plugin-svelte and you should not be using cjs or require synchronous setup to work if you are building a framework. We will not make any guarantees for vite-plugin-svelte to avoid async setup steps in the future.

Vite is also going to remove its explicit cjs support at some point (originally slated for vite6, but postponed https://github.com/vitejs/vite/discussions/13928 ).

ricardo-devis-agullo:
Yeah, that's fine, the fact that my setup requires to be synchronous is because its an older framework and doing changes to that amounts to great deal of pain, so if CJS support gets explicitly removed we will just stick with Vite 6 + Svelte 5 which is fine. Thanks for the response, and I guess we can close this for now!

PR #1113, opened 2025-04-22
fix: tell vite not to optimize clsx

dominikg:
alternative to #1067 

clsx has a single file esm export so does not benefit from being optimized. In theory vite can do better itself here but to help svelte users experiencing this behavior we can add it to optimizeDeps.exclude. This works well even for svelte versions that do not use clsx and requires minimal code change.

Users who want clsx "optimized" can put it in optimizeDeps.include to force it - which will also prevent the dev server restart.


dominikg:
well, vite import analysis didn't like this at all :see_no_evil:

PR #1123, opened 2025-05-30
fix(optimizeDeps): add svelte > clsx to avoid devserver restart

dominikg:
alternative to #1067 only adding it when clsx is a dep of svelte and as a subpath optimization.


dominikg:
messed up the branch, retry

PR #1129, opened 2025-06-03
drop node 18 and update exports syntax

dominikg:
using `default` instead of `import` is possible now since you can use require thanks to "require-esm" support in the node versions we bumped to

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -181,31 +181,16 @@ You should not use prebundleSvelteLibraries during build or for ssr, disable one
 
 You really shouldn't. Svelte and Vite are esm first and the ecosystem is moving away from commonjs and so should you. Consider migrating to esm.
 
-In case you have to, use dynamic import to load vite-plugin-svelte's esm code from cjs like this:
+In case you have to, you can rely on node's "require esm" feature available in v20.19+
 
-```diff
+```js
 // vite.config.cjs
 const { defineConfig } = require('vite');
-- const { svelte } = require('@sveltejs/vite-plugin-svelte');
-module.exports = defineConfig(async ({ command, mode }) => {
-+ const { svelte } = await import('@sveltejs/vite-plugin-svelte');
-  return {plugins:[svelte()]}
-}
-```
+const { svelte, vitePreprocess } = require('@sveltejs/vite-plugin-svelte');
 
-And for `vitePreprocess` you have to set up a lazy promise as top-level-await doesn't work for esm imports in cjs:
-
-```diff
-- const {vitePreprocess} = require('@sveltejs/vite-plugin-svelte')
-+ const vitePreprocess = import('@sveltejs/vite-plugin-svelte').then(m => m.vitePreprocess())
-
-module.exports = {
--    preprocess: vitePreprocess()
-+    preprocess: {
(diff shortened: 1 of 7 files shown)


docs/faq.md
```diff
@@ -181,31 +181,16 @@ You should not use prebundleSvelteLibraries during build or for ssr, disable one
 
 You really shouldn't. Svelte and Vite are esm first and the ecosystem is moving away from commonjs and so should you. Consider migrating to esm.
 
-In case you have to, use dynamic import to load vite-plugin-svelte's esm code from cjs like this:
+In case you have to, you can rely on node's "require esm" feature available in v20.19+
```
eltigerchino, line 184:
```suggestion
In case you have to, you can rely on node's "require esm" feature available in v20.17+
```
According to https://nodejs.org/api/modules.html#loading-ecmascript-modules-using-require this was also backported to v20.17.0. Not sure if it's also in v20.18.0 though since that's not explicitly mentioned

packages/vite-plugin-svelte-inspector/package.json
```diff
@@ -23,7 +21,7 @@
     "generate:types": "dts-buddy -m \"@sveltejs/vite-plugin-svelte-inspector:src/public.d.ts\""
   },
   "engines": {
-    "node": "^18.0.0 || ^20.0.0 || >=22"
+    "node": "^20.19 || ^22.12 || >=24"
```
eltigerchino, line 24:
```suggestion
    "node": "^20.17 || ^22 || >=24"
```
Same as https://github.com/sveltejs/vite-plugin-svelte/pull/1129/files#r2127851453

packages/vite-plugin-svelte/package.json
```diff
@@ -23,7 +21,7 @@
     "generate:types": "dts-buddy -m \"@sveltejs/vite-plugin-svelte:src/public.d.ts\""
   },
   "engines": {
-    "node": "^18.0.0 || ^20.0.0 || >=22"
+    "node": "^20.19 || ^22.12 || >=24"
```
eltigerchino, line 24:
```suggestion
    "node": "^20.17 || ^22 || >=24"
```
Same as https://github.com/sveltejs/vite-plugin-svelte/pull/1129/files#r2127851453

docs/faq.md
dominikg: in 20.17 the feature was behind a flag, it was 20.19 that enabled it by default: https://nodejs.org/en/blog/release/v20.19.0

package.json
```diff
@@ -66,7 +66,7 @@
   "packageManager": "pnpm@10.11.0",
   "engines": {
     "pnpm": "^10.2.0",
-    "node": "^20.18 || ^22 || >=24"
+    "node": "^20.19 || ^22.12 || >=24"
```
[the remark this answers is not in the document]
dominikg: same as above, 22.12 was where it was enabled by default https://nodejs.org/en/blog/release/v22.12.0

docs/faq.md
eltigerchino: Ah right. I misunderstood. My bad.

eltigerchino (approved):

PR #1135, opened 2025-06-06
feat: experimental rolldown-vite support

### LINKED ISSUE #1117: vite-rolldown support
### Describe the problem

Get it passing the ecosystem CI

replace esbuildOptions and transformWithEsbuild and other esbuild stuff in vite-plugin-svelte with the oxc/rolldown equivalent

### Alternatives considered

n/a

### Importance

nice to have

dominikg:
fixes #1117 

the goal is that users can swap in npm:rolldown-vite for vite, vite-plugin-svelte detects it and it *just works*

- [x] add to test matrix
- [x] add detection
- [x] update config
- [x] transformWithEsbuild to transformWithOxc
- [x] esbuild optimizer plugin to rolldown optimizer

optional:
- [ ] run build with both and diff output, make sure that there are no hidden changes

### DIFF
--- a/packages/vite-plugin-svelte/src/utils/optimizer-plugins.js
+++ b/packages/vite-plugin-svelte/src/utils/optimizer-plugins.js
@@ -1,7 +1,7 @@
 import { readFileSync } from 'node:fs';
 import * as svelte from 'svelte/compiler';
 import { log } from './log.js';
-import { toESBuildError } from './error.js';
+import { toESBuildError, toRollupError } from './error.js';
 import { safeBase64Hash } from './hash.js';
 import { normalize } from './id.js';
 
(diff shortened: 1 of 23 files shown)


packages/vite-plugin-svelte/src/utils/optimizer-plugins.js
```diff
@@ -1,17 +1,33 @@
...
 /**
  * @typedef {NonNullable<import('vite').DepOptimizationOptions['esbuildOptions']>} EsbuildOptions
  * @typedef {NonNullable<EsbuildOptions['plugins']>[number]} EsbuildPlugin
  */
-
-export const facadeEsbuildSveltePluginName = 'vite-plugin-svelte:facade';
-export const facadeEsbuildSvelteModulePluginName = 'vite-plugin-svelte-module:facade';
+/*
+ * @typedef {NonNullable<import('vite').DepOptimizationOptions['rollupOptions']>} RollupOptions
+ * @typedef {NonNullable<RollupOptions['plugins']>[number]} RolldownPlugin
```
dominikg, line 14: even if i were to change `vite` in line 13 to `rolldown-vite` rollupOptions.plugins would not be detected here.

so below is a shim that works for our use.

packages/vite-plugin-svelte/src/utils/optimizer-plugins.js
```diff
@@ -49,11 +68,63 @@ export function esbuildSveltePlugin(options) {
...
+				id: { include: [/^[^?#]+\.svelte(?:[?#]|$)/], exclude: [/^virtual-module:/] },
+				code: { exclude: [/(?:import|export)[^"\n]+"virtual-module:/] }
+			},
+			/**
+			 * @param {string} code
+			 * @param {string} filename
+			 */
+			async handler(code, filename) {
+				try {
+					return await compileSvelte(options, { filename, code }, statsCollection);
```
dominikg, line 105: instead of esbuild plugin that returns sourcemaps inline with a comment, this returns code,map.

packages/vite-plugin-svelte/src/utils/optimizer-plugins.js
```diff
@@ -155,11 +230,63 @@ export function esbuildSvelteModulePlugin(options) {
...
+			) {
+				delete plugin.transform;
+				delete plugin.buildStart;
+				delete plugin.buildEnd;
+			}
+		},
+		transform: {
+			filter: {
+				// TODO: remove excludes once above options hook works
+				id: { include: [/^[^?#]+\.svelte\.js(?:[?#]|$)/], exclude: [/^virtual-module:/] },
```
dummdidumm, line 258: This should also include `.ts` I think?

dummdidumm (commented):
I don't _really_ understand what's going on but it looks mostly good to me

packages/vite-plugin-svelte/src/utils/optimizer-plugins.js
dominikg: thanks!! used ` * @typedef {import('vite').Rollup.Plugin} RollupPlugin`

PR #1145, opened 2025-06-22
refactor: modular plugins

dominikg:
this grew rather large but sharing for better discussion

goals:
- split into multiple plugins with clear responsibilities
- separate plugins for preprocess and compile
- fully embrace/adopt environment api
- improve filter use to ensure less work is done in rolldown-vite
- move plugins out of "pre" phase where possible 

note: tests currently fail because raw&direct&custom query support hasn't been adapted to the preprocess/compile split

todo: delete dead code in utilities and move functions into plugins where feasible


packages/vite-plugin-svelte/src/plugins/compile.js
```diff
@@ -0,0 +1,166 @@
...
+					const svelteMeta = this.getModuleInfo(id)?.meta?.svelte;
+					compileData = await compileSvelte(svelteRequest, code, options, svelteMeta?.preprocessed);
+				} catch (e) {
+					cache.setError(svelteRequest, e);
+					throw toRollupError(e, options);
+				}
+				if (compileData.compiled?.warnings) {
+					logCompilerWarnings(svelteRequest, compileData.compiled.warnings, options);
+				}
+				if (svelteRequest.raw) {
```
dominikg, line 48: this block of code is for an experimental feature returning data for import queries like `App.svelte?raw&svelte&type=script`, however that feature has remained unused for a long time.

with the split between preprocess & compile it got more complicated to support it so i propose to remove the feature entirely and only keep basic "raw" support from vite that returns the non-transformed file.

packages/vite-plugin-svelte/src/plugins/load-custom.js
```diff
@@ -0,0 +1,42 @@
...
+
+		load: {
+			//filter: is set in configResolved
+			async handler(id) {
+				const config = this.environment.config;
+				const ssr = config.consumer === 'server';
+				const svelteRequest = api.idParser(id, ssr);
+				if (svelteRequest) {
+					const { filename, raw, query } = svelteRequest;
+					if (!query.url && (raw || config.assetsInclude(filename))) {
```
dominikg, line 29: the check for `raw` here is needed for the custom import query support

packages/vite-plugin-svelte/src/plugins/hot-update.js
```diff
@@ -0,0 +1,180 @@
...
+	if (code == null) {
+		return code;
+	}
+
+	return (
+		code
+			// svelte5 add_location line numbers argument
+			.replace(/(\$\.add_locations\(.*), \[(\[[.[\], \d]+])]/g, '$1, []')
+			// vite import analysis timestamp queries
+			.replace(/[?&]t=\d+/g, '')
```
dominikg, line 178: is `\d+` ok or should it be more targeted eg `\d{13}[&\b]` ?

packages/vite-plugin-svelte/src/plugins/load-custom.js
dominikg: removed custom import queries

.changeset/fair-pets-crash.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sveltejs/vite-plugin-svelte': major
+---
+
+Remove experimental "advanced raw queries" feature. Basic `File.svelte?raw` is still supported.
```
eltigerchino, line 5: Does this mean we should remove vite_css_regex from https://github.com/sveltejs/kit/blob/fb88ff7f2eeeec850130808e79e40191f9b273cb/packages/kit/src/exports/vite/dev/index.js#L247 ?

packages/vite-plugin-svelte-inspector/src/index.js
```diff
@@ -69,42 +73,53 @@ export function svelteInspector(options) {
...
-			if (importee.startsWith('virtual:svelte-inspector-options')) {
-				return importee;
-			} else if (importee.startsWith('virtual:svelte-inspector-path:')) {
-				return importee.replace('virtual:svelte-inspector-path:', inspectorPath);
+		resolveId: {
+			filter: {
+				id: /^virtual:svelte-inspector-/
+			},
+			async handler(importee, _, options) {
+				if (options?.ssr || disabled) {
```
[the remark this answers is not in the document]
dominikg: done

packages/vite-plugin-svelte/src/plugins/hot-update.js
dominikg: updated the regex and docs

packages/vite-plugin-svelte/src/plugins/preprocess.js
```diff
@@ -0,0 +1,133 @@
...
+			if (typeof preprocessed?.map === 'object') {
+				mapToRelative(preprocessed?.map, filename);
+			}
+			return /** @type {import('../types/compile.d.ts').PreprocessTransformOutput} */ {
+				code: preprocessed.code,
+				// @ts-expect-error
+				map: preprocessed.map,
+				meta: {
+					svelte: {
+						preprocessed
```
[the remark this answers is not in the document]
dominikg: thing is that dependencies are really a preprocessor thing, the compiler always only looks at the input svelte code and assumes it to come from one file (except the input sourcemap can reference multiple files but the compiler doesn't use that for changing the output in any way)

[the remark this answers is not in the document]
eltigerchino: Oh interesting. So it should really be in preprocess plugin, that way if it were ever removed, you wouldn't have to fiddle with the compile file?

PR #1154, opened 2025-07-02
refactor: caching

dominikg:
changes:
- move handling of preprocessor dependencies into preprocess plugin
- remove stats for preprocess (that never worked)
- instead of a separate cache for js+css, attach the css output to js module and grab it from moduleinfo in load
- pass input sourcemap into svelte.compile instead of cached preprocessor sourcemap that could be missing transforms


packages/vite-plugin-svelte/src/plugins/compile.js
```diff
@@ -32,37 +31,30 @@ export function compile(api) {
...
-				const cache = api.getEnvironmentCache(this);
 				let compileData;
 				try {
-					const svelteMeta = this.getModuleInfo(id)?.meta?.svelte;
-					compileData = await compileSvelte(svelteRequest, code, options, svelteMeta?.preprocessed);
+					compileData = await compileSvelte(
+						svelteRequest,
+						code,
+						options,
+						this.getCombinedSourcemap()
```
dominikg, line 40: instead of passing in sourcemap from preprocessed that could be outdated because another transform changed the code in between the 2 plugins, use getCombinedSourceMap()  this has the added benefit that we don't have to pass preprocessed around in meta

packages/vite-plugin-svelte/src/utils/preprocess.js
```diff
@@ -80,52 +79,3 @@ export function addExtraPreprocessors(options, config) {
...
-	const selfIdx = [];
-	const normalizedFullFilename = normalizePath(filename);
-	const normalizedDeps = dependencies.map(normalizePath);
-	for (let i = 0; i < normalizedDeps.length; i++) {
-		if (normalizedDeps[i] === normalizedFullFilename) {
-			selfIdx.push(i);
-		}
-	}
-	const hasSelfDependency = selfIdx.length > 0;
-	if (hasSelfDependency) {
```
dominikg, line 107: this was an issue with svelte-preprocess and has been fixed in 2023, we don't need it anymore

packages/vite-plugin-svelte/src/utils/preprocess.js
```diff
@@ -80,52 +79,3 @@ export function addExtraPreprocessors(options, config) {
...
-		warnings.push({
-			code: 'vite-plugin-svelte-preprocess-depends-on-self',
-			message:
-				'svelte.preprocess returned this file as a dependency of itself. This can be caused by an invalid configuration or importing generated code that depends on .svelte files (eg. tailwind base css)',
-			filename
-		});
-	}
-
-	if (dependencies.length > 10) {
```
dominikg, line 116: this was an issue when users put tailwind base rules into svelte style blocks, but that is no longer required with tailwind@4.

existing projects have seen this warning plenty of times and either dealt with it or accepted the penalty. new projects are unlikely to use tailwind@3 in this fashion. 

so we don't need this anymore either

eltigerchino (approved):
Thank you!

Issue #1121, opened 2025-05-23
unprocessed `.svelte` files in transitive dependencies within a monorepo


AdrianGonz97:
### Describe the bug

It seems that `vite-plugin-svelte` doesn't process the `.svelte` files from transitive dependencies when those files come from the package's `devDependencies`, which arguably feels _a bit_ unexpected in the context of a monorepo with private local packages.

In [this example](https://github.com/AdrianGonz97/unresolvable-dev-deps-svelte-monorepo/blob/main/ui/src/components/button/button.svelte), we're importing a svelte component from `@lucide/svelte` (which is installed as dev dep) in the `ui` package for a `Button` component that is then [used in the site `app`](https://github.com/AdrianGonz97/unresolvable-dev-deps-svelte-monorepo/blob/main/app/src/routes/%2Bpage.svelte).

The idea behind this monorepo setup is to have a local `ui` package that isn't necessarily published to npm, so the dependency-type for each dependency of that package 'shouldn't really matter' as the dependency will always be resolvable. 

One solution (if not, _the_ solution) would be to move the `@lucide/svelte` dependency in `ui` to `dependencies`. However, I think it's worth questioning whether this should be necessary at all if the dependency is locally resolvable.

### Reproduction URL

https://github.com/AdrianGonz97/unresolvable-dev-deps-svelte-monorepo

### Reproduction

1. clone https://github.com/AdrianGonz97/unresolvable-dev-deps-svelte-monorepo
2. run `pnpm i && pnpm dev`
3. navigate to `http://localhost:5173`

### Logs

```shell
11:56:17 AM [vite] (ssr) Error when evaluating SSR module /src/routes/+page.svelte: Unknown file extension ".svelte" for /home/koala/Reproductions/unresolvable-dev-deps-svelte-monorepo/node_modules/.pnpm/@lucide+svelte@0.511.0_svelte@5.33.1/node_modules/@lucide/svelte/dist/icons/bubbles.svelte
      at Object.getFileProtocolModuleFormat [as file:] (node:internal/modules/esm/get_format:219:9)
      at defaultGetFormat (node:internal/modules/esm/get_format:245:36)
      at defaultLoad (node:internal/modules/esm/load:120:22)
      at async ModuleLoader.loadAndTranslate (node:internal/modules/esm/loader:580:32)
[...]

PR #1155, opened 2025-07-03
fix: crawl local workspace private packages devDependencies

dominikg:
for generating vite config ssr.noExternal

this improves it for people running into #1121

### DIFF
--- a/packages/vite-plugin-svelte/package.json
+++ b/packages/vite-plugin-svelte/package.json
@@ -44,7 +44,7 @@
     "deepmerge": "^4.3.1",
     "kleur": "^4.1.5",
     "magic-string": "^0.30.17",
-    "vitefu": "^1.1.0"
+    "vitefu": "^1.1.1"
   },
   "peerDependencies": {
     "svelte": "^5.0.0",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -658,8 +658,8 @@ importers:
         specifier: ^0.30.17
         version: 0.30.17
       vitefu:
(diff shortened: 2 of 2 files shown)


AdrianGonz97:
this is now working perfectly in the [updated reproduction](https://github.com/AdrianGonz97/unresolvable-dev-deps-svelte-monorepo/tree/pr-1155-working) 🎉

eltigerchino:
Is it worth adding a test for this with a setup similar to the reproduction?

packages/e2e-tests/kit-node/__tests__/kit.spec.ts
```diff
@@ -307,13 +310,22 @@ describe('kit-node', () => {
...
-						'@sveltejs/adapter-node'
+						'@sveltejs/adapter-node',
+						'e2e-test-dep-svelte-nested-workspace-devdep',
+						'e2e-test-dep-svelte-simple'
 					],
 					`ssr.noExternal in ${filename}`
 				);
+				const expectedExcludes = ['@sveltejs/kit', '$app', '$env'];
+				if (!isServe) {
+					expectedExcludes.push(
```
dominikg, line 321: optimizeDeps.exclude during build isn't really needed anymore as newer vite no longer supports it, but a change for that would have to come in vitefu and that still supports older versions that have it.

i wonder if we should omit optimizeDeps config in the plugin instead of generating values that are ignored.

dominikg:
> Is it worth adding a test for this with a setup similar to the reproduction?

turns out it was, found a windows path handling issue :heart:

Issue #1165, opened 2025-07-09
enhance enhanceCompileError to report when no svelte config was found


maxlath:
### Describe the problem

Hi!

I have been struggling with migrating a Svelte project from webpack to vite, and finally came to a breakthrough when I realized my `svelte.config.js` had to be moved to a different directory (`./app` instead of `.`) to be detected by `vite`/`vite-plugin-svelte`

----

For the sake of making that issue easier to find by people going on the same road, here is the kind of error message I was getting with my config in the wrong place:

```
error during build:
[vite-plugin-svelte] [plugin vite-plugin-svelte] app/path/to/component.svelte (89:2): app/path/to/component.svelte:89:2 Expected a valid CSS identifier
https://svelte.dev/e/css_expected_identifier

- Did you forget to add a scss preprocessor? See https://github.com/sveltejs/vite-plugin-svelte/blob/main/docs/preprocess.md for more information.
```

### Describe the proposed solution

`enhanceCompileError` was reporting that no preprocessor was set, but would have saved me a good amount of banging my head on the wall had it told me that no config was found at all.

Also logging the path where the config file would be expected would be great (my breakthrough came after I ran `vite --debug` and realized that `vite` was looking for env files in `./app`)

### Alternatives considered

Giving up and becoming a goat keeper

### Importance

would make my life easier

maxlath:
I now realize I could have found that info by running `export DEBUG=vite-plugin-svelte:* ; vite`. Is there maybe a way to enable `vite-plugin-svelte` debug logs when running `vite --debug`?

maxlath:
I tried to see if that was possible to detect the `--debug` flag from the plugin, but couldn't make it work

```js
export function svelte(inlineOptions) {
	// Not using node:util parseArgs, as that makes vite exit without doing anything
	const args = process.argv.slice(2);
	let arg, debugFlag, debugNamespace;
	while (arg = args.shift()) {
		if (arg === '-d' || arg === '--debug') {
			debugFlag = true;
			const nextArg = args[0];
			if (nextArg && !nextArg.startsWith('-')) {
				debugNamespace = nextArg;
			}
			break;
		}
	}
	if (process.env.DEBUG != null || debugFlag) {
		log.setLevel('debug');
	}
        log.debug('hello?', undefined, 'config') // Doesn't log anything
```

dominikg:
[...]

PR #1166, opened 2025-07-09
fix: do not wrongly suggest that a lang attribute is missing when finding comments

maxlath:
Additionally to the issue reported in #1165, `enhanceCompileError` was wrongly reporting that some of my `<style>` blocks were missing a `lang` attribute, due to the `styleRe` also matching comment block. In This PR, I thus suggest to not match comment blocks for that purpose.


dominikg:
as far as i remember this regex is coming from https://github.com/sveltejs/svelte/blob/9926347ad9dbdd0f3324d5538e25dcb7f5e442f8/packages/svelte/src/compiler/preprocess/index.js#L257  and the part you removed here is used to consume comments to avoid processing style blocks inside comments.

do you have a reproduction where a commented style is matched and processed by this? kind of reluctant to diverge from sveltes regex here (ultimately i'd love to get rid of it entirely, it is much too complex)

maxlath:
I pushed a fixup commit 69a2e0b that would preserver the regex, and here is a demo of the issue https://github.com/maxlath/vite-plugin-svelte-demo/

maxlath:
I ran `pnpm changeset`, squashed the fixup commits, and force pushed

dominikg (approved):

PR #1185, opened 2025-07-28
feat: enable optimizer for server environments during dev

dominikg:


packages/vite-plugin-svelte/src/plugins/setup-optimizer.js
```diff
@@ -94,54 +77,72 @@ export function setupOptimizer(api) {
...
- * @param {EsbuildPlugin} plugin
- * @param {import('../types/options.d.ts').ResolvedOptions} options
+ * @param {import('../types/plugin-api.d.ts').PluginAPI} api
+ * @param {'server'|'client'} consumer
+ * @param {boolean} components
+ * @return {EsbuildPlugin}
  */
-function patchESBuildOptimizerPlugin(plugin, options) {
-	const components = plugin.name === optimizeSveltePluginName;
+function eSBuildOptimizerPlugin(api, consumer, components) {
```
eltigerchino, line 85:
```suggestion
function esBuildOptimizerPlugin(api, consumer, components) {
```

packages/vite-plugin-svelte/src/plugins/setup-optimizer.js
```diff
@@ -43,43 +44,25 @@ export function setupOptimizer(api) {
...
+						rolldownOptimizerPlugin(api, consumer, true),
+						rolldownOptimizerPlugin(api, consumer, false)
 					]
 				};
 			} else {
 				optimizeDeps.esbuildOptions = {
 					plugins: [
-						{ name: optimizeSveltePluginName, setup: () => {} },
-						{ name: optimizeSvelteModulePluginName, setup: () => {} }
+						eSBuildOptimizerPlugin(api, consumer, true),
```
eltigerchino, line 57:
```suggestion
						esbuildOptimizerPlugin(api, consumer, true),
```

packages/vite-plugin-svelte/src/plugins/setup-optimizer.js
```diff
@@ -43,43 +44,25 @@ export function setupOptimizer(api) {
...
+						eSBuildOptimizerPlugin(api, consumer, false)
```
eltigerchino, line 58:
```suggestion
						esbuildOptimizerPlugin(api, consumer, false)
```

packages/vite-plugin-svelte/src/plugins/setup-optimizer.js
```diff
@@ -32,7 +32,8 @@ export function setupOptimizer(api) {
 	return {
 		name: 'vite-plugin-svelte:setup-optimizer',
 		apply: 'serve',
-		config() {
+		configEnvironment(_name, config) {
+			const consumer = config.consumer ?? 'client';
```
eltigerchino, line 36: `configEnvironment` doesn't apply the default consumer here https://github.com/vitejs/vite/blob/557f797d29422027e8c451ca50dd84bf8c41b5f0/packages/vite/src/node/config.ts#L787-L789 so this always evaluates to "client" for `@cloudflare/vite-plugin` since it doesn't explicitly set the `consumer` option but instead relies on Vite setting it to the default of "server" when the options are fully resolved (after this hook?).

dominikg: great catch, i guess we'll have to replicate the code from vite.

packages/vite-plugin-svelte/src/plugins/configure.js
```diff
@@ -87,6 +94,12 @@ export function configure(api, inlineOptions) {
 			ensureConfigEnvironmentConditions(name, config, opts);
 			// @ts-expect-error the function above should make `resolve.conditions` non-nullable
 			config.resolve.conditions.push('svelte');
+			if (config.consumer === 'server' && optimizeDeps !== undefined) {
+				// optimizeDeps is not inherited by server environments so return it here
+				return {
+					optimizeDeps
```
dominikg, line 100: not entirely happy with this for 2 reasons

1) this configuration part should probably be moved out of this plugin and into the setup-optimizer plugin, but that requires a bit of refactoring and maybe running vitefu twice.

2) it replicates the same optimizeDeps for server and client env, but i am not sure yet if the same rules apply

eltigerchino: > it replicates the same optimizeDeps for server and client env, but i am not sure yet if the same rules apply

I was thinking the same thing. Would it be enough to just pass `extensions` and `plugins` so that esbuild knows how to handle Svelte modules? Then the user can decide what to include/exclude for optimizing themselves.

packages/vite-plugin-svelte/src/plugins/setup-optimizer.js
```diff
@@ -94,54 +78,72 @@ export function setupOptimizer(api) {
...
-		build.onEnd(() => {
-			statsCollection?.finish();
-		});
+			build.onLoad({ filter }, async ({ path: filename }) => {
+				const code = readFileSync(filename, 'utf8');
+				try {
+					const result = await compileFn(
+						api.options,
+						{ filename, code },
+						generate,
```
eltigerchino, line 111:
```suggestion
						generate: consumer,
```

Issue #1188, opened 2025-08-05
Watch mode generates corrupt CSS on second build


Fd929c2CE5fA:
### Describe the bug

When using vite api build with watch mode, even if saved without changes, the generated **second or later** automatic build will produce an incorrect CSS file, it contains irrelevant svelte (script/html) code.

After testing, the `"@sveltejs/vite-plugin-svelte": "5.1.1"` version works correctly and the `"@sveltejs/vite-plugin-svelte": "6.0.0"` and `"@sveltejs/vite-plugin-svelte": "6.1.0"` versions can reproduce the error.

That is, the issue exist in plugin version > 6.

### Reproduction URL

https://stackblitz.com/edit/vitejs-vite-q6tkjnjy?file=src%2Flib%2FCounter.svelte&terminal=dev

### Reproduction

1. The first automatic build (`node watch.js`) after opening the stackblitz link generates the correct CSS:
<details>
<summary>dist/assets/index-8HoovIC0.css</summary>

```css
:root {
  font-family: system-ui, Avenir, Helvetica, Arial, sans-serif;
  line-height: 1.5;
  font-weight: 400;

  color-scheme: light dark;
  color: rgba(255, 255, 255, 0.87);
  background-color: #242424;

  font-synthesis: none;
  text-rendering: optimizeLegibility;
  -webkit-font-smoothing: antialiased;
  -moz-osx-font-smoothing: grayscale;
}

a {
  font-weight: 500;
  color: #646cff;
  text-decoration: inherit;
}
a:hover {
  color: #535bf2;
}

body {
  margin: 0;
  display: flex;
  place-items: center;
  min-width: 320px;
  min-height: 100vh;
}

h1 {
  font-size: 3.2em;
  line-height: 1.1;
}

.card {
  padding: 2em;
}

#app {
  max-width: 1280px;
  margin: 0 auto;
  padding: 2rem;
  text-align: center;
}

button {
  border-radius: 8px;
  border: 1px solid transparent;
  padding: 0.6em 1.2em;
  font-size: 1em;
  font-weight: 500;
  font-family: inherit;
  background-color: #1a1a1a;
  cursor: pointer;
  transition: border-color 0.25s;
}
button:hover {
  border-color: #646cff;
}
button:focus,
button:focus-visible {
  outline: 4px auto -webkit-focus-ring-color;
}

@media (prefers-color-scheme: light) {
  :root {
    color: #213547;
    background-color: #ffffff;
  }
[...]

PR #1189, opened 2025-08-06
fix `build --watch` 

### LINKED ISSUE #1186: Version 6 breaks`build --watch` throwing [commonjs] [postcss] Unknown word xxx
### Describe the bug

Hi, after a long debugging session I think I found the root cause of a compilation error that is breaking my pipeline.
After upgrading from `@sveltejs/vite-plugin-svelte` from `5.1.0` to `6.0.0-next.1` (or newer) the build in watch mode has started to throw this error:
```
[commonjs] [postcss] <path.svelte> Unknown word Table <== a type imported inside script lang ts
file: <path.svelte>?svelte&type=style&lang.css:2:13
```

for example the mentioned file <path.svelte> starts with:
```
<script lang="ts">
    import { Table } from '@sveltestrap/sveltestrap';
```

I just need to transpile SCSS in svelte components, I don't need (or at least explicitly in my configuration) `postcss` etc.
```
<style lang="scss">
</style>
```

The build initially works fine, it's only after I change (and save) a file that this error is thrown. Btw a normal build works fine.

I'm using an `inlined` config, we can say a bit custom, so for sure I encountered a weird edge case.
Any tips on what is going on?

Many thanks for any help.

### Reproduction URL

https://vite.new/svelte

### Reproduction

```json
// package.json

{
  "scripts": {
    "v3-watch-all": "cross-env NODE_ENV=development MODE=development node ./App-v3/pipeline/vite.build.mjs",
  }
}
```

```js
// ./App-v3/pipeline/vite.build.mjs

import path from 'node:path'
import fs from 'node:fs';
import { build } from 'vite'
import { svelte, vitePreprocess } from '@sveltejs/vite-plugin-svelte';

/* 
 * VITE
 * command: vite/vite dev/vite serve => 'serve', vite build => 'build'
 * mode: vite dev => 'development', vite build => 'production'
*/

if (!process.env.NODE_ENV) {
    throw new Error('process.env.NODE_ENV not set [production|development]')
}

if (!process.env.MODE) {
    throw new Error('process.env.MODE not set [production|development]')
}

const basePath = process.cwd();
const isProduction = process.env.NODE_ENV === 'production';
const isWatch = process.env.MODE === 'development';
const outDir = path.resolve(basePath, 'wwwroot/dist-v3/build');

console.log('Using configurations', {
    'MODE:': process.env.MODE,
    'NODE_ENV': process.env.NODE_ENV,
    basePath,
    isProduction,
    isWatch,
    outDir
});

const apps = [
    { root: path.resolve(basePath, './Views/AdminV3/'), alias: 'adminv3', entry: './Views/AdminV3/AdminV3.entry.ts' },
    { root: path.resolve(basePath, './Views/AdminPvcV3/'), alias: 'adminpvcv3', entry: './Views/AdminPvcV3/AdminPvcV3.entry.ts' },
    { root: path.resolve(basePath, './Views/InspectionOverviewV3/'), alias: 'inspectionoverviewv3', entry: './Views/InspectionOverviewV3/InspectionOverviewV3.entry.ts' },
    { root: path.resolve(basePath, './Views/InspectionPvcV3/'), alias: 'inspectionpvcv3', entry: './Views/InspectionPvcV3/InspectionPvcV3.entry.ts' },
    { root: path.resolve(basePath, './Views/InspectionV3/'), alias: 'inspectionv3', entry: './Views/InspectionV3/InspectionV3.entry.ts' }
];

// to be able to generate separated js and css assets, we need to build each app separately.
const builds = apps.map(app => {
    console.log('Building app', { app });

    return build({
        root: app.root,
        appType: 'custom',
        build: {
            // https://rollupjs.org/configuration-options/#watch
            watch: isWatch ? {
                buildDelay: 100,
                clearScreen: true,
                exclude: 'node_modules/**'
            } : undefined,

            // https://rollupjs.org/configuration-options/
            rollupOptions: {
                input: {
                    [app.alias]: app.entry
                },
                output: {
                    entryFileNames: '[name].js',
                    chunkFileNames: '[name].js',
                    assetFileNames: `${app.alias}.[ext]`,
                },

                onwarn(warning, warn) {
                    if (warning.code === 'INVALID_ANNOTATION' && warning.id.includes('node_modules')) {
                        return;
                    }
                    warn(warning);
                },
            },

            // we don't have a "public" dir, we use fs.cpSync(...) instead
            copyPublicDir: false,

            outDir: outDir,
            sourcemap: true,
            modulePreload: {
                polyfill: true
            },
            emptyOutDir: false,
            cssCodeSplit: false,
            minify: isProduction,
            reportCompressedSize: false,

            // default 500
            chunkSizeWarningLimit: 2500
        },
        plugins: [
            svelte({
                emitCss: true,
                experimental: {
                    disableSvelteResolveWarnings: true
                },
                preprocess: [vitePreprocess({ script: false, style: true })],
                compilerOptions: {
                    compatibility: {
                        componentApi: 4
                    },
                    css: 'external',

                    // enable run-time checks when not in production
                    dev: !isProduction,
                },

                // Warnings are normally passed straight to Rollup. You can optionally handle them here, for example to squelch warnings with a particular code
                onwarn: (warning, handler) => {
                    if (warning.filename?.endsWith('/SimpleAutocomplete.svelte')) {
                        return;
                    }

                    // let Rollup handle all other warnings normally
                    handler && handler(warning);
                },
            })
        ],

    })
});

// top level async function is not supported by "concurrently", the vite process exits immediately
async function buildApps() {
    for (let build of builds) {
        await build;
    }
}

try {
    console.log(`Preparing ${outDir}`);

    try {
        // to avoid concurrency issues, the folder could be cleaned from a custom npm script in package.json, "v[x]-clean".
        fs.mkdirSync(outDir, { recursive: true });
    } catch (err) {
        console.error(err)
    }

    console.log('Building svelte apps');

    buildApps();

} catch (e) {
    console.error(e);
    process.exit(1);
}

/*
 * NB. Do not return an exit code, for example via `process.exit(0)`, otherwise
 * "concurrently" will not wait for the script completion while in watch mode.
 */
```

### Logs

```shell

```

### System Info

```shell
System:
    OS: Windows 11 10.0.26100
    CPU: (20) x64 12th Gen Intel(R) Core(TM) i7-12700H
    Memory: 26.62 GB / 63.68 GB
  Binaries:
    Node: 22.17.0 - C:\Program Files\nodejs\node.EXE
    npm: 11.4.2 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: Chromium (131.0.2903.86)
    Internet Explorer: 11.0.26100.1882
  npmPackages:
    @sveltejs/vite-plugin-svelte: ^6.0.0-next.1 => 6.1.0
    svelte: ^5.37.1 => 5.37.1
    vite: ^7.0.6 => 7.0.6
```

dominikg:
- [x] add test mode for `build --watch`
- [x] add test setup that  works with page refresh after watch rebuild
- [x] fix issues uncovered

fixes #1186 and #1188 

due to `getModuleInfo` in load hook not returning cached data for unchanged modules, compiled css was not loaded and instead vites own fallback load plugin returned the uncompiled file content.

To fix it, this PR caches the css output separately and returns it on rebuild.


package.json
```diff
@@ -3,10 +3,11 @@
   "private": true,
   "type": "module",
   "scripts": {
-    "test": "run-s -c test:unit \"test:build {@}\" \"test:serve {@}\" --",
+    "test": "run-s -c test:unit \"test:build {@}\" \"test:serve {@}\" \"test:build:watch {@}\" --",
     "test:unit": "vitest run",
     "test:serve": "vitest run -c vitest.config.e2e.ts",
     "test:build": "cross-env TEST_BUILD=1 vitest run -c vitest.config.e2e.ts",
+    "test:build:watch": "cross-env TEST_BUILD_WATCH=1 vitest run -c vitest.config.e2e.ts",
```
dominikg, line 10: `build --watch` was not tested at all before but 2 recent issues highlighted that we can't just depend on it being tested by vite itself

packages/e2e-tests/_test_dependencies/vite-plugins/index.js
```diff
@@ -1,5 +1,15 @@
 import path from 'node:path';
 import fs from 'node:fs';
+import MagicString from 'magic-string';
+
+function replaceWithSourceMap(code, value, replacement) {
```
dominikg, line 5: the missing sourcemap lead to error logs, add them to allow easier debugging and assertions that no unexpected errors happened

packages/e2e-tests/e2e-server.js
```diff
@@ -46,7 +46,33 @@ async function startedOnPort(serverProcess, port, timeout) {
...
+	const timerPromise = new Promise(
+		(_, reject) =>
+			(id = setTimeout(() => {
+				reject(`timeout for server start after ${timeout}`);
+			}, timeout))
+	);
+	const startedPromise = new Promise((resolve, reject) => {
+		stdoutListener = (data) => {
+			const str = data.toString();
+			const match = str.match(/built in \d+ms\./);
```
dominikg, line 61: initial build of `build --watch` logs "build in 123ms." at the end. wait for that to ensure we don't try to load the preview before that is done

packages/e2e-tests/testUtils.ts
```diff
@@ -204,6 +204,29 @@ export async function editFileAndWaitForHmrComplete(file, replacer, fileUpdateTo
...
 
+export async function editFileAndWaitForBuildWatchComplete(file, replacer) {
+	const newContent = editFile(file, replacer);
+
+	try {
+		await waitForBuildWatchAndPageReload(hmrUpdateTimeout);
+	} catch (e) {
+		const maxTries = isCI && isWin ? 3 : 1;
+		let lastErr;
+		for (let i = 1; i <= maxTries; i++) {
```
dominikg, line 215: not sure if we need this, unlike hmr where edits can happen super fast there is a build step in between here so it should work the first time always

packages/e2e-tests/testUtils.ts
```diff
@@ -231,11 +254,35 @@ export async function saveScreenshot(name: string) {
...
 
+export async function waitForBuildWatchAndPageReload(timeout = 10000) {
+	const logs = e2eServer.logs.watch.out;
+	const startIdx = logs.length;
+	let timeleft = timeout;
+	const pollInterval = 50;
+	let completed = false;
+	while (timeleft > 0) {
+		await sleep(pollInterval);
+		if (logs.some((text, i) => i > startIdx && text.match(/\d+ modules transformed\./))) {
```
dominikg, line 270: build watch logs the number of transformed modules each time a watch change triggered it.

packages/e2e-tests/vite-ssr-esm/package.json
```diff
@@ -5,9 +5,7 @@
   "type": "module",
   "scripts": {
     "dev": "node server",
-    "build": "run-s build:client build:server",
-    "build:client": "vite build --ssrManifest .vite/ssr-manifest.json --outDir dist/client",
-    "build:server": "vite build --ssr src/entry-server.js --outDir dist/server",
+    "build": "vite build --app",
```
dominikg, line 8: for `--watch` to work swiched this over to environment api builder, the previous scripts would not work as they relied on being called in sequence.

packages/vite-plugin-svelte/src/plugins/load-compiled-css.js
```diff
@@ -37,6 +54,8 @@ export function loadCompiledCss(api) {
 					}
 					css.moduleType = 'css';
 					return css;
+				} else {
+					log.warn(`failed to load virtual css module ${id}`, undefined, 'load');
```
dominikg, line 58: with this fix, this should not happen anymore, but we want to know if it does because that would be another issue.

bluwy (approved):

Issue #1192, opened 2025-08-08
v6 prints errors during `vite dev`: `[postcss] Unknown word xxx`


echocrow:
### Describe the bug

Originally reported here: https://github.com/sveltejs/vite-plugin-svelte/issues/1186#issuecomment-3160944151

After upgrading to v6, running `vite dev` can cause excessive error logs in the terminal in the form of:
```
[vite] (ssr) Error when evaluating SSR module <file-path>: [postcss] <file-path:line:col>: Unknown word <word>
  Plugin: vite:css
  ...
```
(exact plugin being reported may vary)

This issue is present both under vite `^v6.3.5` and vite `^v7.0.0`.

Downgrading `@sveltejs/vite-plugin-svelte` back to `v5.0.3`, this issue is not present.

### Reproduction URL

https://stackblitz.com/edit/sveltejs-vite-plugin-svelte-vite-ssr-bug?file=src%2Froutes%2F%2Bpage.svelte

### Reproduction

Some pieces that were required to replicate this:
- a postcss config with at least one plugin
- a browser-only lazy-imported component that uses postcss

Given the repro:
1. Run `pnpm dev`
2. Then click "reload" in the preview pane (or browser tab during local dev)

(in our production repo, hitting refresh is not needed for the error to be logged; unsure why this minimal repo does need the manual refresh)

### Logs

```shell
vite:cache [memory] __sveltekit/paths +6s
  vite:resolve 0.04ms /src/routes/Foobar.svelte?inline&svelte&type=style&lang.css -> /home/projects/sveltejs-vite-plugin-svelte-vite-ssr-bug/src/routes/Foobar.svelte?inline&svelte&type=style&lang.css +6s
  vite:load 3.42ms [fs] /src/routes/Foobar.svelte?inline&svelte&type=style&lang.css +6s
2:04:33 PM [vite] (ssr) Error when evaluating SSR module /src/routes/Foobar.svelte?inline&svelte&type=style&lang.css: [postcss] /home/projects/sveltejs-vite-plugin-svelte-vite-ssr-bug/src/routes/Foobar.svelte?inline&svelte&type=style&lang.css:7:255: Unknown word '{'
  Plugin: vite:css
  File: /home/projects/sveltejs-vite-plugin-svelte-vite-ssr-bug/src/routes/Foobar.svelte?inline&svelte&type=style&lang.css:7:254
[...]

PR #1194, opened 2025-08-12
fix: add testcase for failing dev reload

dominikg:
see https://github.com/sveltejs/vite-plugin-svelte/issues/1192

this testcase adds a noop postcss plugin and a dynamically imported component using postcss.

On page reload the dev server logs an error.


packages/e2e-tests/kit-node/package.json
```diff
@@ -3,8 +3,8 @@
   "version": "0.0.0",
   "private": true,
   "scripts": {
-    "dev": "vite dev",
-    "build": "vite build",
+    "dev": "svelte-kit sync && vite dev",
```
dominikg, line 6: without this the server log could contain a warning "no .svelte-kit/tsconfig.json" found. Not needed in users projects but this simple call avoids it

dummdidumm (approved):

PR #1201, opened 2025-08-18
chore(docs): restructure docs so they can be ingested by svelte.dev

dominikg:
This moves the docs markdown files into a structure parsable by svelte.dev generator and adds symlinks in the old places so that existing links keep working.

### DIFF
--- a/documentation/docs/index.md
+++ b/documentation/docs/index.md
@@ -1,0 +1,3 @@
+---
+title: vite-plugin-svelte
+---
--- a/documentation/index.md
+++ b/documentation/index.md
@@ -1,3 +1,0 @@
----
-Title: vite-plugin-svelte
----


documentation/docs/10-plugin/10-config.md
eltigerchino, line 1: Is it worth thinking about de-duplicating the config option info between this file and in the JSDocs? (generating the docs from the JSDoc types like we do in Svelte and SvelteKit)

dominikg: how does that work and can it be added to a pre-commit hook + ci check?

eltigerchino: I think you'll have to ask Elliot who made new changes to the svelte.dev bot https://discordapp.com/channels/457912077277855764/1385015784564330589/1409640631160012810

For how the content looks like, there's an example in https://github.com/sveltejs/kit/blob/main/documentation/docs/98-reference/50-configuration.md

PR #1207, opened 2025-08-28
feat(rolldown-vite): enable optimization.inlineConst by default

dominikg:
This is going to help with treeshaking and esm-env. 

We have to wait for a new release of rolldown (1.0.0-beta.35) and rolldown-vite using that before merging.


paoloricciuti (approved):
Looks good to me, just left a comment but if you feel it's not the case feel free to merge!

PR #1227, opened 2025-09-24
Harden publish

dominikg:
- introduce composite action to reduce repeated steps
- move call to publish into a separate job with
  - an assigned environment `release` that prevents this job from executing on other branches
  - minimal permissions
  - no pnpm install

### DIFF
--- a/.github/actions/setup-node/action.yml
+++ b/.github/actions/setup-node/action.yml
@@ -12,7 +12,7 @@ runs:
   using: composite
   steps:
     # the following 2 steps are a replacement for pnpms own setup action that contains a lot of code
-    - uses: actions/setup-node@a0853c24544627f65ddf259abe73b1d18a591444 # v5.0.0
+    - uses: actions/setup-node@395ad3262231945c25e8478fd5baf05154b1d79f # v6.1.0
       with:
         node-version: ${{inputs.node-version}}
         package-manager-cache: false
@@ -22,7 +22,7 @@ runs:
         PNPM_VER=$(jq -r '.packageManager | if .[0:5] == "pnpm@" then .[5:] else "packageManager in package.json does not start with pnpm@\n" | halt_error(1)  end' package.json)
         echo installing pnpm version $PNPM_VER
         npm i --ignore-scripts -g pnpm@$PNPM_VER
-    - uses: actions/setup-node@a0853c24544627f65ddf259abe73b1d18a591444 # v5.0.0
+    - uses: actions/setup-node@395ad3262231945c25e8478fd5baf05154b1d79f # v6.1.0
       if: ${{inputs.install-deps == 'true'}}
       with:
         node-version: ${{inputs.node-version}}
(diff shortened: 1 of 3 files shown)


.github/actions/setup-node/action.yml
```diff
@@ -0,0 +1,33 @@
...
+  node-version:
+    description: 'node version'
+    default: '24'
+  install-deps:
+    description: 'install deps'
+    default: 'true'
+
+runs:
+  using: composite
+  steps:
```
dominikg, line 13: these steps are required in all of our jobs, this makes it easier to update and reduced code duplication

.github/workflows/release.yml
```diff
@@ -4,57 +4,110 @@ on:
...
+      - name: Harden the runner
+        uses: step-security/harden-runner@f4a75cfd619ee5ce8d5b864b0d183aff3c69b55a # v2.13.1
+        with:
+          egress-policy: block
+          allowed-endpoints: >
+            api.github.com:443
+            github.com:443
+            release-assets.githubusercontent.com:443
+            registry.npmjs.org:443
+            *.sigstore.dev:443
```
dominikg, line 85: this extra endpoint is required for oidc publish

.github/workflows/release.yml
```diff
@@ -4,57 +4,110 @@ on:
...
+          # generate "--filter package1 --filter package2"
+          PUBLISH_PACKAGES_FILTER=$(echo '${{needs.changesets.outputs.publishedPackages}}' | jq -r '.[] | " --filter " + "\"" + .name+ "\""' | xargs)
+
+          if [[ -f .changeset/pre.json ]]; then
+            TAG=$(jq -r '.tag' .changeset/pre.json)
+          else
+            TAG=latest
+          fi
+
+          if [[ "$GIT_STATUS" != "" ]]; then
```
dominikg, line 106: ensure that git state is still clean.This would only fail if an action used by this job or the runner itself was compromised, but better safe than sorry.

.github/workflows/ci.yml
```diff
@@ -100,24 +89,12 @@ jobs:
...
-          node-version: ${{ matrix.node }}
-          package-manager-cache: false # pnpm is not installed yet
-      - name: install pnpm
-        shell: bash
-        run: |
-          PNPM_VER=$(jq -r '.packageManager | if .[0:5] == "pnpm@" then .[5:] else "packageManager in package.json does not start with pnpm@\n" | halt_error(1)  end' package.json)
-          echo installing pnpm version $PNPM_VER
-          npm i -g pnpm@$PNPM_VER
-      - uses: actions/setup-node@v5
+          persist-credentials: 'false'
```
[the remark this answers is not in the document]
dominikg: i think its pretty self-explanatory. we don't need persisted git credentials but the action defaults to true.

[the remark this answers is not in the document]
benmccann: yeah, I guess you're right. such a terrible default :stuck_out_tongue_closed_eyes:  I guess that's the part that really confused me, but that's an issue for github

dominikg:
back to draft after #1263 has been merged

Issue #1238, opened 2025-11-11
`optimizeDeps.rollupOptions` will be removed


sapphi-red:
### Describe the problem

`optimizeDeps.rollupOptions` is used here. I added a runtime deprecation in the latest version of rolldown-vite and I'm planning to remove this in the next version of rolldown-vite.
https://github.com/sveltejs/vite-plugin-svelte/blob/752752460151eee9dea4f41648538998cb9abeec/packages/vite-plugin-svelte/src/plugins/setup-optimizer.js#L46
(Note: `build.rollupOptions` is also deprecated but would still be there)

### Describe the proposed solution

Replace it with `optimizeDeps.rolldownOptions`.

### Alternatives considered

Not sure

### Importance

would make my life easier

dominikg:
vite-plugin-svelte@6.2.2 uses rolldownOptions.

A new beta for vite-plugin-svelte with official support for vite8 is starting soon.

PR #1249, opened 2025-12-27
Update peer dependency for vite to include version 8

rubenfiszel:


rubenfiszel:
Possibily this could/should be addressed: https://github.com/sveltejs/vite-plugin-svelte/issues/1238

dominikg:
Thanks but while vite8 is still in beta and support is experimental it isn't in peer Dependencies on purpose. The plan is to start a new major soonish that adds vite8 and drops support for older versions. 

This involves a bit more work so I'll close here.

PR #1257, opened 2026-01-07
fix: process style modules at the end during HMR

ocavue:
During the development of https://github.com/withastro/astro/pull/14821, I noticed that sometimes HMR doesn't work. After some debugging, here is the reason in my opinion:

1. When the `<style>` tag in a `.svelte` file changes, [`hotUpdate.handler`](https://github.com/sveltejs/vite-plugin-svelte/blob/83513a8226d22b5f8c23cf30d63c27ae38c2194c/packages/vite-plugin-svelte/src/plugins/hot-update.js#L80) is called. [`ctx.modules`](https://github.com/sveltejs/vite-plugin-svelte/blob/83513a8226d22b5f8c23cf30d63c27ae38c2194c/packages/vite-plugin-svelte/src/plugins/hot-update.js#L83) contains two modules: `Counter.svelte?svelte&type=style&lang.css` and `Counter.svelte`. Notice the order here. 
2. [`await transformRequest(mod.url)`](https://github.com/sveltejs/vite-plugin-svelte/blob/83513a8226d22b5f8c23cf30d63c27ae38c2194c/packages/vite-plugin-svelte/src/plugins/hot-update.js#L102) is called first for `Counter.svelte?svelte&type=style&lang.css`.
3. The Vite plugin [`loadCompiledCss`](https://github.com/sveltejs/vite-plugin-svelte/blob/83513a8226d22b5f8c23cf30d63c27ae38c2194c/packages/vite-plugin-svelte/src/plugins/load-compiled-css.js#L10)'s `load.handler` method is called, and a [cached CSS](https://github.com/sveltejs/vite-plugin-svelte/blob/83513a8226d22b5f8c23cf30d63c27ae38c2194c/packages/vite-plugin-svelte/src/plugins/load-compiled-css.js#L39) is returned. 
4. Since the CSS for `Counter.svelte?svelte&type=style&lang.css` is from the previous cache, the module `Counter.svelte?svelte&type=style&lang.css` is marked as [not affected](https://github.com/sveltejs/vite-plugin-svelte/blob/83513a8226d22b5f8c23cf30d63c27ae38c2194c/packages/vite-plugin-svelte/src/plugins/hot-update.js#L104-L105).
5. [`await transformRequest(mod.url)`](https://github.com/sveltejs/vite-plugin-svelte/blob/83513a8226d22b5f8c23cf30d63c27ae38c2194c/packages/vite-plugin-svelte/src/plugins/hot-update.js#L102) is called for `Counter.svelte`.
6. The Vite plugin [compile](https://github.com/sveltejs/vite-plugin-svelte/blob/83513a8226d22b5f8c23cf30d63c27ae38c2194c/packages/vite-plugin-svelte/src/plugins/compile.js#L8)'s `transform.handler` method is called. In this step, the `.svelte` file is actually handled by the Svelte compiler, and the cached CSS is updated.
7. Since the JS part of this `.svelte` file doesn't change, the module `Counter.svelte` is marked as not affected too.
8. Both modules are not affected, thus HMR is not triggered.

This PR fixes the issue by processing `Counter.svelte` first, which triggers the Svelte compiler and updates the cached CSS. Only after that, it processes `Counter.svelte?svelte&type=style&lang.css`.

---

I don't know how to test this patch in this pull request, though. While running [hmr.spec.ts](https://github.com/sveltejs/vite-plugin-svelte/blob/83513a8226d22b5f8c23cf30d63c27ae38c2194c/packages/e2e-tests/hmr/__tests__/hmr.spec.ts), the [`ctx.modules`](https://github.com/sveltejs/vite-plugin-svelte/blob/83513a8226d22b5f8c23cf30d63c27ae38c2194c/packages/vite-plugin-svelte/src/plugins/hot-update.js#L83) are in the order [`HmrTest.svelte`, `HmrTest.svelte?svelte&type=style&lang.css`]. This order won't cause the HMR issue I described above. I suspect that the different order of `ctx.modules` has something to do with the fact that `astro` is an SSR framework.

### DIFF
--- a/packages/vite-plugin-svelte/src/plugins/hot-update.js
+++ b/packages/vite-plugin-svelte/src/plugins/hot-update.js
@@ -81,15 +81,28 @@ export function hotUpdate(api) {
 				const svelteRequest = idParser(ctx.file, false, ctx.timestamp);
 				if (svelteRequest) {
 					const { modules } = ctx;
-					const svelteModules = [];
+					const svelteJsModules = [];
+					const svelteCssModules = [];
 					const nonSvelteModules = [];
 					for (const mod of modules) {
 						if (transformResultCache.has(mod.id)) {
-							svelteModules.push(mod);
+							if (mod.id && SVELTE_VIRTUAL_STYLE_ID_REGEX.test(mod.id)) {
+								svelteCssModules.push(mod);
+							} else {
+								svelteJsModules.push(mod);
+							}
 						} else {
 							nonSvelteModules.push(mod);
 						}
 					}
+
+					// Always process Svelte components first, then styles. This
+					// ensures that when we call
+					// `this.environment.transformRequest(mod.url)` later for
+					// style modules in this function, the Svelte components
+					// have already been processed and the cached styles are up
+					// to date.
+					const svelteModules = [...svelteJsModules, ...svelteCssModules];
 
 					if (svelteModules.length === 0) {
 						return; // nothing to do for us


packages/vite-plugin-svelte/src/plugins/hot-update.js
```diff
@@ -81,16 +81,29 @@ export function hotUpdate(api) {
 				const svelteRequest = idParser(ctx.file, false, ctx.timestamp);
 				if (svelteRequest) {
 					const { modules } = ctx;
-					const svelteModules = [];
+					const svelteJsModules = [];
```
ocavue: I'm not sure if `svelteJsModules` is a good variable name or not.

PR #1258, opened 2026-01-07
fix: transform Svelte component first during HMR

ocavue:
Another attempt for https://github.com/sveltejs/vite-plugin-svelte/pull/1257

Closes https://github.com/sveltejs/vite-plugin-svelte/pull/1257

~~Closes #<span></span>1032~~

### DIFF
--- a/.changeset/fair-eyes-burn.md
+++ b/.changeset/fair-eyes-burn.md
@@ -2,4 +2,4 @@
 '@sveltejs/vite-plugin-svelte': patch
 ---
 
-Ensure that the corresponding Svelte component is compiled prior to transforming the CSS within it.
+fix(hmr): ensure that svelte files are recompiled and updated css modules returned correctly when reported out of order


dominikg:
Is there a way where the css-id only case could now lead to a js update being missed? if transformRequest leads to a change in the js module but the module id is not part of the original modules list it would not be added.

packages/vite-plugin-svelte/__tests__/hot-update.spec.js
```diff
@@ -0,0 +1,212 @@
...
+`;
+
+describe('hmr', () => {
+	const svelteRelativePath = 'apple.svelte';
+	const styleRelativePath = svelteRelativePath + '?svelte&type=style&lang.css';
+	const svelteAbsolutePath = path.resolve(fixtureDir, svelteRelativePath);
+	const styleAbsolutePath = path.resolve(fixtureDir, styleRelativePath);
+
+	beforeEach(async () => {
+		fs.writeFileSync(svelteAbsolutePath, initialCode);
```
dominikg: files modified by test cases must be gitignored (e2e setups are copied to a temp dir for this reason)

also a bit worried about drift in the runner setup code, but the approach is interesting esp. because it does not require a full e2e app

ocavue:
If the test fails on Windows, it is likely related to differences in file path formats. For example, a Rollup module ID is not necessarily the same as a file system path, especially on Windows.

However, since the test also fails on Linux, the issue may (also?) relate to the behavior of the file system watcher, which would be more difficult to debug.

Given this, I will migrate the test from unit tests to e2e tests to ensure it works across different environments.

dominikg:
thanks a lot! Do you want to add the changeset or should I?

ocavue:
> had to rerun node20-linux once but the fail was a file system fluke in another test that you didn't touch.

Adding [`retry`](https://vitest.dev/config/retry.html#retry) in `vitest.config.e2e.ts` might improve such cases.

dominikg (approved):

dominikg:
published as 6.2.4

PR #1302, opened 2026-03-13
chore: bump to vite 8

teemingc:


packages/vite-plugin-svelte/package.json
```diff
@@ -45,12 +45,12 @@
   },
   "peerDependencies": {
     "svelte": "^5.46.4",
-    "vite": "^8.0.0-beta.7 || ^8.0.0"
+    "vite": "^8.0.0 || ^8.0.0"
```
Rich-Harris:
```suggestion
    "vite": "^8.0.0"
```

Rich-Harris (commented):
does this require a breaking change? someone could be using this with a Vite 8 beta and presumably if they updated v-p-s the peerDependencies constraints would no longer be satisfied.

on the other hand, that would seem very persnickety and annoying.

I think this needs a changeset?

teemingc:
Whoops, the peerDep should be unchanged. I accidentally changed it during a search and replace. It's reverted now so it should only be devDeps being updated

packages/vite-plugin-svelte/package.json
```diff
@@ -45,12 +45,12 @@
   },
   "peerDependencies": {
     "svelte": "^5.46.4",
-    "vite": "^8.0.0-beta.7 || ^8.0.0"
+    "vite": "^8.0.0"
```
teemingc:
```suggestion
    "vite": "^8.0.0-beta.7 || ^8.0.0"
```

teemingc:
Closing in favour of https://github.com/sveltejs/vite-plugin-svelte/pull/1327

PR #1318, opened 2026-04-10
feat: pass environment name to dynamicCompileOptions

tomyan:
## Summary

Passes the Vite environment name (e.g. `client`, `ssr`) to the `dynamicCompileOptions` callback, enabling per-environment compiler configuration.

## Motivation

With Vite's environment API, projects can define custom environments beyond `client` and `ssr`. The Svelte compiler's `experimental.customRenderer` option (not yet merged — see sveltejs/svelte#18042) will enable rendering to non-DOM targets, but there's currently no way to configure this per-environment through `vite-plugin-svelte` — `generate` is controlled by the plugin based on the consumer type, and `dynamicCompileOptions` doesn't know which environment it's compiling for.

This change adds the environment name to the `dynamicCompileOptions` callback data, allowing users to return different compiler options per environment:

```js
svelte({
    dynamicCompileOptions({ environment }) {
        if (environment === 'terminal') {
            return {
                generate: 'client',
                css: 'external',
                experimental: { customRenderer: 'my-custom-renderer' },
            };
        }
    }
})
```

## Changes

- Pass `environment` name from the compile plugin's transform handler to `compileSvelte`
- Include `environment` in the `dynamicCompileOptions` callback data object
- Add `environment?: string` to the TypeScript type definition
- Two new unit tests: one verifying the environment name is passed, one verifying `generate` can be overridden per environment

## Backwards compatibility

The `environment` field is optional in the type definition and is a new addition to the callback data object. Existing `dynamicCompileOptions` callbacks that don't destructure `environment` are unaffected. The `compileSvelte` function's new `environment` parameter is also optional and defaults to `undefined`.


teemingc:
cc: @paoloricciuti since you wrote the custom renderer

packages/vite-plugin-svelte/src/plugins/compile.js
```diff
@@ -42,7 +42,8 @@ export function compile(api) {
 						svelteRequest,
 						code,
 						options,
-						this.getCombinedSourcemap()
+						this.getCombinedSourcemap(),
+						this.environment.name
```
teemingc, line 46: Why not pass the entire environment object? 🤔 I'd imagine `environment.consumer` would be equally useful

teemingc:
AFAIK you should be able to achieve this with the `perEnvironmentPlugin` helper exported by Vite as documented in https://vite.dev/guide/api-environment-plugins#per-environment-plugins
```js
import { perEnvironmentPlugin, defineConfig } from 'vite';
import svelte from '@sveltejs/vite-plugin-svelte';

export default defineConfig({
  plugins: [
    perEnvironmentPlugin('per-environment-plugin', (environment) =>
      svelte({
        dynamicCompileOptions() {
          if (environment.name === 'terminal') {
            return { /* ... */ }
          }
        }
      }),
    ),
  ],
});
```

There's a note about how it only runs after `configResolved` so I'm worried this might cause the plugin to not configure the vite config as it should. But the docs also mention "Environment plugins resolution may be moved before configResolved in the future.". So it'd be better if they just did that to fix it

PR #1324, opened 2026-04-15
fix: inspector rtl host styles

mustafa0x:
inspector looks bad in rtl sites:
image

## Summary
- add an RTL-specific override for `#svelte-inspector-host`
- force the inspector host to stay `ltr` and use `system-ui` so the overlay UI renders correctly in RTL documents

## Why
The inspector UI should not inherit RTL directionality from the page root, otherwise its own controls and text layout can become awkward or incorrect.

## Validation
- inspected the resulting diff
- attempted local prettier/eslint checks, but the binaries are not available in this checkout (`pnpm exec prettier` / `pnpm exec eslint` failed with command not found)

### DIFF
--- a/.changeset/tasty-eggs-stay.md
+++ b/.changeset/tasty-eggs-stay.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/vite-plugin-svelte': patch
+---
+
+fix: enforce ltr styles for inspector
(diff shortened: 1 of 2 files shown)


Rich-Harris (approved):
thank you! I removed the `font: system-ui` change, not because I disagree with it (the default serif looks atrocious) but because the design changes needed are bigger than this and warrant their own PR

PR #1326, opened 2026-04-21
fix: restore value imports stripped by oxc in script preprocessing

### LINKED ISSUE #1313: Runtime error with Vite 8 and vitePreprocess( { script: true}) in Threlte project with @threlte/extras import
### Describe the bug

Hi all,

When I tried to upgrade to Vite 8 to leverage its new build speed increases, I ran into an issue with a production Capacitor/Threlte project where I was getting runtime errors on build due to an erroneous undefined reference to Box3. When I ran `vite build` with minify: false, I could see that one of the references to Box3 wasn't get renamed like the others:

```js
new Box3$1();
new Color$1();
from_html(`<!> <!>`, 1);
from_html(`<!> <!>`, 1);
new Box3();
from_html(`<!> <!>`, 1);
```

I added the `--sourcemap` option to my build and it showed that the errors were coming from `Resize.svelte`, a component that I do not use:

```svelte
<script
  lang="ts"
  module
>
  // reuse box for all instances
  const _box = new Box3() // <-- produces Uncaught (in promise) ReferenceError: Box3 is not defined
</script>

<script lang="ts">
  import {
    isInstanceOf,
    observe,
    T,
    useStage,
    useTask,
    useThrelte,
    type Plugin
  } from '@threlte/core'
  import { Box3, Group } from 'three'
  import InjectPlugin from '../InjectPlugin/InjectPlugin.svelte'
  import type { ResizeProps } from './types.js'

// ...
```

After thorough investigation and troubleshooting, I discovered that the issue resulted from the project's use of Typescript enums. Building with Typescript enums causes a failure in the vite build that says to add `preprocess: vitePreprocess({ script: true })` to `svelte.config.js`.

When this script step was combined with a few basic imports from `@threlte` and `three` , it led to a runtime error that prevented the app from functioning:

```js
 import { OrbitControls, RoundedBoxGeometry } from '@threlte/extras'
 import { T } from "@threlte/core";
 import { Box3, Box3Helper, Mesh } from "three";
```

I was able to reproduce this by installing Threlte on a fresh project, adding some basic imports from @threlte/extras, and building with the `preprocess: vitePreprocess({ script: true })` step.

The workaround was to change all Typescript enums to `as const` and remove the preprocess step from `svelte.config.js`. This removes all runtime errors.

Please let me know if there's any other information I can provide to help resolve this.

### Reproduction URL

https://github.com/fortserious/vitepreprocess-bug

### Reproduction

Run `pnpm install`
Run `pnpm preview`

You will get a 500 error that says Box3 is undefined.

### Logs

```shell
ReferenceError: Box3 is not defined
    <anonymous> Immutable
```

### System Info

```shell
System:
    OS: Windows 10 10.0.19045
    CPU: (12) x64 Intel(R) Core(TM) i7-8850H CPU @ 2.60GHz
    Memory: 8.51 GB / 31.83 GB
  Binaries:
    Node: 22.14.0 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.21 - C:\Users\ross\AppData\Roaming\npm\yarn.CMD     
    npm: 10.9.2 - C:\Program Files\nodejs\npm.CMD
    pnpm: 9.12.3 - C:\Users\ross\AppData\Roaming\npm\pnpm.CMD      
    Deno: 1.46.2 - C:\Users\ross\.deno\bin\deno.EXE
  Browsers:
    Chrome: 146.0.7680.165
    Edge: Chromium (140.0.3485.54)
    Firefox: 148.0.2 - C:\Program Files\Mozilla Firefox\firefox.exe
    Internet Explorer: 11.0.19041.5794
  npmPackages:
    @sveltejs/adapter-auto: ^7.0.1 => 7.0.1 
    @sveltejs/adapter-static: ^3.0.10 => 3.0.10 
    @sveltejs/kit: ^2.55.0 => 2.55.0 
    @sveltejs/vite-plugin-svelte: ^7.0.0 => 7.0.0 
    svelte: ^5.55.0 => 5.55.0 
    vite: ^8.0.3 => 8.0.3
```

martinmunillas:
When vitePreprocess uses transformWithOxc to strip TypeScript from <script lang="ts"> blocks, oxc removes value imports that appear unused within the script. However, these imports may be referenced in the Svelte template (e.g., imported components used only in markup). This causes ReferenceError during SSR in the development server.

Detect stripped value imports by comparing identifiers before and after the oxc transform, and re-prepend any that were removed. Type-only imports are correctly left stripped.

I found this migrating to vite 8 on a svelte app. It's a company software so I can't share the code, I can try to get a minimum reproducible example if needed.

### DIFF
--- a/packages/vite-plugin-svelte/src/preprocess.js
+++ b/packages/vite-plugin-svelte/src/preprocess.js
@@ -50,24 +50,130 @@ function viteScript() {
 			const { code, map } = await transformWithOxc(content, filename, {
 				lang,
 				target: 'esnext'
-				// TODO, how to pass tsconfig compilerOptions (or not needed as config is loaded for file
-				/*tsconfigRaw: {
-					compilerOptions: {
-						// svelte typescript needs this flag to work with type imports
-						importsNotUsedAsValues: 'preserve',
-						preserveValueImports: true
-					}
-				}*/
 			});
+
+			// oxc strips imports it considers unused, but some may be referenced in the Svelte
+			// template which oxc cannot see. Detect stripped value imports and re-add them.
+			const restoredCode = restoreStrippedImports(content, code);
 
 			mapToRelative(map, filename);
 
 			return {
-				code,
+				code: restoredCode,
 				map
 			};
 		}
 	};
+}
+
+/**
+ * @typedef {{ kind: 'default' | 'named' | 'namespace', local: string, imported?: string }} ImportedId
+ */
+
+/**
+ * Extracts imported value identifiers from an import statement.
+ * Skips type-only specifiers (e.g., `type Foo` inside `import { type Foo, Bar }`).
+ * @param {string} importStatement
+ * @returns {{ ids: ImportedId[], source: string } | null}
+ */
+function getImportedIdentifiers(importStatement) {
+	// skip `import type { ... }` or `import type Foo` entirely
+	if (/\bimport\s+type\b/.test(importStatement)) return null;
+
+	const sourceMatch = importStatement.match(/from\s+['"]([^'"]+)['"]/);
+	if (!sourceMatch) return null;
+	const source = sourceMatch[1];
+
+	/** @type {ImportedId[]} */
+	const ids = [];
+
+	// default import: import Foo from '...'
+	const defaultMatch = importStatement.match(/import\s+([A-Za-z_$][\w$]*)\s+from/);
+	if (defaultMatch) ids.push({ kind: 'default', local: defaultMatch[1] });
+
+	// named imports: import { a, b as c, type D } from '...'
+	const namedMatch = importStatement.match(/import\s*(?:[A-Za-z_$][\w$]*\s*,\s*)?\{([^}]+)\}/);
+	if (namedMatch) {
+		for (const specifier of namedMatch[1].split(',')) {
+			const trimmed = specifier.trim();
+			if (!trimmed) continue;
+			// skip type-only specifiers: `type Foo` or `type Foo as Bar`
+			if (/^type\s+/.test(trimmed)) continue;
+			const asMatch = trimmed.match(/^(\S+)\s+as\s+(\S+)$/);
+			if (asMatch) {
+				ids.push({ kind: 'named', imported: asMatch[1], local: asMatch[2] });
+			} else {
+				ids.push({ kind: 'named', local: trimmed });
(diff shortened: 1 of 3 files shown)


martinmunillas:
hi @Rich-Harris , not sure if this is a right approach or if you see any issues but happy to receive feedback

packages/vite-plugin-svelte/src/preprocess.js
```diff
@@ -49,15 +49,12 @@ function viteScript() {
 			const lang = /** @type {'ts'} */ (attributes.lang);
 			const { code, map } = await transformWithOxc(content, filename, {
 				lang,
-				target: 'esnext'
-				// TODO, how to pass tsconfig compilerOptions (or not needed as config is loaded for file
```
teemingc, line 53: I wonder if we should set `tsconfig: true` so that the tsconfig is auto-detected, resolving this comment.

teemingc (changes requested):
Thanks! Can you add a changeset for this?

packages/vite-plugin-svelte/src/preprocess.js
martinmunillas: I'm guessing that might not be a good idea as then specific tsconfigs may mess with onlyRemoveTypeImports, for example verbatimModuleSyntax. I'm not sure how oxc would handle that.

Isn't it's whole job just to strip TypeScript types so Svelte can compile the script? I think adding tsconfig: true will just add unpredictability.

I might be wrong, let me know what you think.

teemingc: I’m not too sure myself. I’d have to try to get in contact with Dominik to find out what he meant by this comment

martinmunillas:
@teemingc added the changeset, copied it from https://github.com/sveltejs/vite-plugin-svelte/pull/1330/commits/b188f7f3511386af93d2a2aa7e56f6ab08bc620f

packages/vite-plugin-svelte/src/preprocess.js
teemingc: I'm guessing it's to match esbuild's default behaviour of finding the nearest tsconfig https://esbuild.github.io/api/#tsconfig . We can always enable it in another PR if needed

teemingc (approved):
Thanks again!

PR #1328, opened 2026-04-23
feat: enable optimizer for server environments during dev

teemingc:
building on top of https://github.com/sveltejs/vite-plugin-svelte/pull/1185

This PR ensures Svelte works correctly if the Vite dev server is in "full-bundle mode" (`optimizeDeps` on everything) such as when Cloudflare's Vite plugin is added. It ensures the dev server doesn't reload on initial page request and that Svelte is always bundled on the server unless specifically excluded.

The difference between this and the previous PR is an attempt to address Dominik's two concerns:
1. Instead of simply copying the client optimized deps, we are also including server-only Svelte exports
2. Moving the configuration from the setup plugin to the optimize plugin

### DIFF
--- a/.changeset/wet-cats-study.md
+++ b/.changeset/wet-cats-study.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/vite-plugin-svelte': minor
+---
+
+feat: enable optimizer for server environments during dev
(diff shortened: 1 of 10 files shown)


Rich-Harris (approved):
LGTM!

teemingc:
There's a test failing on Windows because of a letter drive difference in the sourcemap path 🤦🏼 

EDIT: fixed in https://github.com/sveltejs/kit/pull/15793

mrkishi:
Seems like it was unintentionally closed by the related PR.

PR #1329, opened 2026-04-24
fix: use correct pnpm catalog syntax

teemingc:
It's `catalog:` not `:catalog` 🤦🏼 this should fix our ecosystem CI failures


dummdidumm (approved):
oops I missed that too 😅

PR #1340, opened 2026-05-06
chore: remove `fs-extra`

teemingc:
This PR removes `fs-extra` in favour of the built-in `node:fs` by following the e18e migration guide e18e.dev/docs/replacements/fs-extra.html


packages/e2e-tests/vitestSetup.ts
```diff
@@ -125,8 +126,10 @@ beforeAll(
 					const segments = file.split(path.sep);
 					return segments.some((segment) => directoriesToIgnore.includes(segment));
 				};
-				await fs.copy(srcDir, tempDir, {
+				// eslint-disable-next-line n/no-unsupported-features/node-builtins -- cp is available in Node 20 and we only use it for test setup anyway
```
teemingc, line 129: This eslint error fires because it's marked as experimental for Node versions below 22. However, it's not gated behind an experimental flag so we can use it while still supporting Node 20. 
https://nodejs.org/api/fs.html#fspromisescpsrc-dest-options

.prettierrc.js
```diff
@@ -27,7 +27,8 @@ export default {
 				'packages/playground/big/src/pages/**', // lots of generated files
 				'packages/e2e-tests/scan-deps/src/Svelte*.svelte', // various syntax tests that require no format
 				'**/.vite-inspect/**',
-				'packages/e2e-tests/_test_dependencies/**/*.svelte' // TODO remove after sourcemap bug is fixed
+				'packages/e2e-tests/_test_dependencies/**/*.svelte', // TODO remove after sourcemap bug is fixed
+				'packages/e2e-tests/*/logs'
```
teemingc, line 31: Avoid running prettier on gitignored logs

package.json
```diff
@@ -31,16 +31,14 @@
     "@sveltejs/eslint-config": "^10.0.0",
     "@sveltejs/kit": "catalog:",
     "@svitejs/changesets-changelog-github-compact": "^1.2.0",
-    "@types/fs-extra": "^11.0.4",
-    "@types/node": "^22.19.1",
+    "@types/node": "^20.19.39",
```
teemingc, line 34: Better to keep use Node 20 types while we still support it. We can probably drop Node 20 support in the next major since Kit 3 supports 22 and above

package.json
```diff
@@ -8,7 +8,7 @@
     "test:serve": "vitest run -c vitest.config.e2e.ts",
     "test:build": "TEST_BUILD=1 vitest run -c vitest.config.e2e.ts",
     "test:build:watch": "TEST_BUILD_WATCH=1 vitest run -c vitest.config.e2e.ts",
-    "check": "run-p -c check:*",
+    "check": "run-p -c 'check:*'",
```
teemingc, line 11: Adding the quotes around the argument helps it run correctly in zsh shells

benmccann (approved):

PR #1341, opened 2026-05-06
chore: replace `execa` with `tinyexec`

teemingc:
This PR follows the e18e recommendation for replacing `execa` with a lighter alternative https://e18e.dev/docs/replacements/execa.html . The recommended packages can be easily compared at https://npmx.dev/compare?packages=execa,tinyexec,nanoexec


packages/e2e-tests/tsconfig.json
```diff
@@ -8,7 +8,9 @@
 		"module": "ESNext",
 		"outDir": "dist",
 		"baseUrl": ".",
+		"checkJs": true,
```
teemingc, line 11: Enables type checking for the `packages/e2e-tests/e2e-server.js` file which helped in migrating towards the `tinyexec` API

packages/e2e-tests/tsconfig.json
```diff
@@ -8,7 +8,9 @@
...
 		"allowJs": true,
+		"noEmit": true,
```
teemingc, line 13: Not sure why this was never added. This tsconfig is only used for IDE type errors anyway. Adding `noEmit` gets rid of the VS Code error saying TypeScript would overwrite existing modules if we ever ran `tsc`

benmccann (approved):

PR #1355, opened 2026-05-30
fix(inspector): activate under Vite+ by matching its client module path

OrientusPrime:
## Problem
With `vitePlugin: { inspector: true }`, the Svelte Inspector doesn't activate under
[Vite+](https://viteplus.dev) — holding the toggle key (default Alt+X) does nothing.
The same config works under stock Vite.

## Cause
The inspector injects its runtime by appending an import to the dev client module,
matched by a path filter in `inspector/index.js`:

    filter: { id: /vite\/dist\/client\/client\.mjs(?:\?|$)/ }

This matches stock Vite's `…/vite/dist/client/client.mjs`. Vite+ resolves its client
to `…/@voidzero-dev/vite-plus-core/dist/vite/client/client.mjs`
(`CLIENT_ENTRY = path.join(VITE_PACKAGE_DIR, "dist/vite/client/client.mjs")`) — the
`vite` and `dist` segments are in the opposite order, so the filter doesn't match and
the loader import is never appended.

The client module still passes through the transform pipeline under Vite+ (only
`vite:dynamic-import-vars` / `vite:asset-import-meta-url` exclude it, as in stock
Vite), so widening the filter is enough.

## Change
Accept both segment orders:

    filter: { id: /(?:vite\/dist|dist\/vite)\/client\/client\.mjs(?:\?|$)/ }

## Verification
On a SvelteKit app running Vite+, `/@vite/client` then contains the injected
`load-inspector` import and the inspector mounts (Alt+X works). Stock Vite still
matches via the `vite/dist` branch.

This looked like the smallest change that covers it, and it held up in testing — but
whether to support Vite+, and how, is your call.

### DIFF
--- a/.changeset/inspector-vite-plus-client.md
+++ b/.changeset/inspector-vite-plus-client.md
@@ -2,4 +2,4 @@
 '@sveltejs/vite-plugin-svelte': patch
 ---
 
-fix: activate the inspector under Vite+ by also matching its dev client module path (`dist/vite/client/client.mjs`)
+fix: ensure the inspector is injected into the client correctly for Vite+ projects
--- a/.changeset/inspector-vite-plus-client.md
+++ b/.changeset/inspector-vite-plus-client.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/vite-plugin-svelte': patch
+---
+
+fix: activate the inspector under Vite+ by also matching its dev client module path (`dist/vite/client/client.mjs`)
--- a/packages/vite-plugin-svelte/src/plugins/inspector/index.js
+++ b/packages/vite-plugin-svelte/src/plugins/inspector/index.js
@@ -115,7 +115,10 @@ export function svelteInspector(api) {
 			}
 		},
 		transform: {
-			filter: { id: /(?:vite\/dist|dist\/vite)\/client\/client\.mjs(?:\?|$)/ },
+			// Vite+ projects install vite@npm:@voidzero-dev/vite-plus-core@latest which
+			// changes the path from `vite/dist/client/client.mjs` to `vite/dist/vite/client/client.mjs`
+			// so we need to also account for the additional `vite` subdirectory
+			filter: { id: /vite\/dist\/(?:vite\/)?client\/client\.mjs(?:\?|$)/ },
 			handler(code) {
 				if (disabled) {
 					return;
--- a/packages/vite-plugin-svelte/src/plugins/inspector/index.js
+++ b/packages/vite-plugin-svelte/src/plugins/inspector/index.js
@@ -115,7 +115,7 @@ export function svelteInspector(api) {
 			}
 		},
 		transform: {
-			filter: { id: /vite\/dist\/client\/client\.mjs(?:\?|$)/ },
+			filter: { id: /(?:vite\/dist|dist\/vite)\/client\/client\.mjs(?:\?|$)/ },
 			handler(code) {
 				if (disabled) {
 					return;


.changeset/inspector-vite-plus-client.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sveltejs/vite-plugin-svelte': patch
+---
+
+fix: activate the inspector under Vite+ by also matching its dev client module path (`dist/vite/client/client.mjs`)
```
teemingc:
```suggestion
fix: ensure the inspector is injected into the client correctly for Vite+ projects
```

packages/vite-plugin-svelte/src/plugins/inspector/index.js
```diff
@@ -115,7 +115,7 @@ export function svelteInspector(api) {
 			}
 		},
 		transform: {
-			filter: { id: /vite\/dist\/client\/client\.mjs(?:\?|$)/ },
+			filter: { id: /(?:vite\/dist|dist\/vite)\/client\/client\.mjs(?:\?|$)/ },
```
teemingc:
```suggestion
			// Vite+ projects install vite@npm:@voidzero-dev/vite-plus-core@latest which
			// changes the path from `vite/dist/client/client.mjs` to `vite/dist/vite/client/client.mjs`
			// so we need to also account for the additional `vite` subdirectory
			filter: { id: /vite\/dist\/(?:vite\/)?client\/client\.mjs(?:\?|$)/ },
```

teemingc (approved):

teemingc:
Thank you!

dexvdev (commented):
The merged filter assumes the id contains vite/dist/vite/client/client.mjs — i.e. that the vite alias directory name survives in the path. That's true for npm/yarn, where vite@npm:@voidzero-dev/vite-plus-core materializes a real node_modules/vite directory. Under pnpm, node_modules/vite is a symlink, and Node's ESM resolver resolves it to the realpath, so VITE_PACKAGE_DIR (derived from import.meta.url) — and therefore CLIENT_ENTRY — ends up as:

…/node_modules/.pnpm/@voidzero-dev+vite-plus-core@0.2.1_…/node_modules/@voidzero-dev/vite-plus-core/dist/vite/client/client.mjs

There is no vite/dist/ substring in that path, so the merged filter never matches and the inspector doesn't activate — same symptom as before the fix.

Verified in a pnpm monorepo on vite-plus@0.2.1:

await import.meta.resolve('vite')
// file:///…/node_modules/.pnpm/@voidzero-dev+vite-plus-core@0.2.1_…/node_modules/@voidzero-dev/vite-plus-core/dist/vite/node/index.js

/vite\/dist\/(?:vite\/)?client\/client\.mjs(?:\?|$)/.test(clientEntry) // false
/(?:vite\/dist|dist\/vite)\/client\/client\.mjs(?:\?|$)/.test(clientEntry)  // true (originally proposed form)

The originally proposed alternation (?:vite\/dist|dist\/vite) covers both segment orders regardless of package manager. Happy to open a follow-up PR if you'd like.

dexvdev:
Following up on my review comment above — confirmed by a live test on a pnpm monorepo (`vite-plus@0.2.1`, pnpm 11.10.0, Node 24.11.0): with unpatched `@sveltejs/vite-plugin-svelte@7.2.0` the inspector never activates (Alt+X does nothing), because the pnpm-realpathed client path `…/@voidzero-dev/vite-plus-core/dist/vite/client/client.mjs` contains no `vite/dist/` substring for the merged filter to match.

Opened #1373 with the one-line fix (making the leading `vite/` optional) plus full details and the test environment.

PR #1357, opened 2026-06-16
chore: migrate to pnpm 11

teemingc:
pnpm 11 only works on node 22 or newer so we can't merge this until the min supported version is 22 or we just drop the 20 tests because most people would have upgraded already

### DIFF
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -16,14 +16,6 @@ minimumReleaseAgeExclude:
   - svelte-check
   - esm-env
 
-onlyBuiltDependencies:
-  - esbuild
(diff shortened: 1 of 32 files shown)


pnpm-workspace.yaml
```diff
@@ -33,12 +25,11 @@ auditConfig:
     - 'GHSA-f886-m6hf-6m8v' # belongs to an optional peer dependency of Vite
 
 overrides:
-  # $ prefixed package names take their version from root package.json - this is like catalogs but without them
```
teemingc, line 36: $ syntax has been deprecated by pnpm in favour of catalog:

PR #1362, opened 2026-06-27
feat(inspector): add a wrapper-chain dropdown to jump to where a component is used

dexvdev:
## What

While the inspector is active, a new key combo — `alt-c` — enters "chain mode":
clicking a node opens a dropdown of its **wrapper chain**, i.e. every ancestor
`<Component>` with the exact `file:line` where it is used. Pick the wrapper you
want and it opens in your editor via the same `/__open-in-editor` call. `alt-x`
and the normal click are unchanged — this is purely additive and opt-in.

## Why

The inspector points at where a DOM element is _defined_. For a primitive reused
across an app (e.g. a shared `<Button>`), every instance resolves to the same
`button.svelte` — never the `<Button …/>` call site in the parent, which is what
you usually want to edit. That call site isn't a DOM node, so neither the element
location nor arrow-key DOM nav can reach it.

## How

Svelte's dev build already records the data: `el.__svelte_meta` carries `.loc`
(the element's own tag) and `.parent` (the dev component stack, whose
`type: 'component'` entries hold `{ file, line, column, componentTag }` for each
call site in the parent template). The dropdown walks that linked list. No new
runtime data and no compiler change.

## Test plan

- `packages/e2e-tests/inspector-vite` now renders `<Counter />` in `App.svelte`;
  a new dev-only e2e enters chain mode, clicks the button, and asserts the
  dropdown lists both `Counter.svelte` (the leaf) and `App.svelte` (the
  `<Counter>` call site).
- `pnpm check:types`, `eslint`, `prettier --check`, and the `inspector-vite` e2e
  (serve + build) all pass.
- Docs (`docs/inspector.md`) and a `minor` changeset are included.

### DIFF
--- a/.changeset/inspector-wrapper-chain.md
+++ b/.changeset/inspector-wrapper-chain.md
@@ -1,0 +1,7 @@
+---
+'@sveltejs/vite-plugin-svelte': minor
+---
+
+feat(inspector): add a wrapper-chain dropdown (alt-c) to jump to where a component is used
+
+While the inspector is active, `alt-c` enters "chain mode": clicking a node opens a dropdown of its wrapper chain — every ancestor `<Component>` with its exact `file:line` call site (walked from Svelte dev's `__svelte_meta.parent`). Hovering a row outlines that component's DOM; clicking opens it via the same `/__open-in-editor` call. `alt-x` keeps its original single-element behavior.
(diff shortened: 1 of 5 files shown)


Rich-Harris:
Thank you. We shouldn't add a new key combo for this — key combos that don't conflict with anything else are extremely hard to find, and often you wouldn't realise you needed to see the full component stack until you had already enabled the inspector.

Instead, it should be a context menu — people already know how to use these, it doesn't require a new key combo, will work better with mobile when we get around to making the inspector work better on mobile, and can be extended with new things in future. Closing this in favour of #1370, which builds on this branch.

PR #1370, opened 2026-07-07
feat: context menu with component stack

Rich-Harris:
Adds the component stack from #1362, but behind a (right-click) context menu rather than a new key combo. Also unifies the styles between the two modes, and gives them a much needed facelift


paoloricciuti:
image

Is it expected that it also shows the generated `root.svelte` component?

Rich-Harris:
err no, I thought I'd squelched that — lemme fix (it'll be hardcoded, so if you customise that directory they'll still peek through)

paoloricciuti (approved):

PR #1383, opened 2026-07-28
fix: stub out exported snippets during dependency scanning

### LINKED ISSUE #1356: Exported snippets from <script module> fail Vite dependency scan when lang="ts" is omitted
### Describe the bug

When exporting a Svelte 5 snippet from `<script module>` without using `lang="ts"`, I get this error:

```
(!) Failed to run dependency scan. Skipping dependency pre-bundling.
Error: Failed to scan for dependencies from entries:
  /src/routes/+layout.svelte
  /src/routes/+page.svelte

  Build failed with 1 error:
  [PARSE_ERROR] Export 'logo' is not defined
     ╭─[ virtual-module:/src/lib/icons.svelte?id=0:2:11 ]
     │
   2 │     export { logo };
     │              ──┬─
     │                ╰───
```

Adding `lang="ts"` to `<script module>` fixes it.

### Reproduction URL

https://github.com/matiadev/sveltekit-snippet-export-error

### Reproduction

1. Clone the repo and run `pnpm install`
2. Run `pnpm dev --force` to see the error
3. Add `lang="ts"` to `<script module>` in `src/lib/icons.svelte`
4. Run `pnpm dev --force` again and it works

### Logs

```shell

```

### System Info

```shell
svelte: 5.56.2
@sveltejs/kit: 2.63.0
@sveltejs/vite-plugin-svelte: 7.1.2
vite: 8.0.16
```

jkbz64:
Fixes #1356

~~Make vite-plugin-svelte compile the whole .svelte component during dependency scanning, so template-declared exports such as top-level snippets can be resolved.~~

Stub out exported snippets during depenendency scanning so that template-declared (non lang="ts") snippet exports can be resolved.

### DIFF
--- a/packages/vite-plugin-svelte/src/plugins/setup-optimizer.js
+++ b/packages/vite-plugin-svelte/src/plugins/setup-optimizer.js
@@ -98,7 +98,21 @@ function rolldownOptimizerPlugin(api, consumer, components) {
 			delete plugin.buildStart;
 			delete plugin.transform;
 			delete plugin.buildEnd;
+			if (components) {
+				plugin.load = {
+					filter: { id: includeRe },
+					async handler(filename) {
+						try {
+							const code = await fs.readFile(filename.replace(/[?#].*$/s, ''), 'utf8');
+							return await compileFn(api.options, { filename, code }, generate, statsCollection);
+						} catch (e) {
+							throw toRollupError(e, api.options);
+						}
+					}
+				};
+			}
 		} else {
+			delete plugin.load;
 			plugin.transform = {
 				filter: { id: includeRe },
 				/**
--- a/packages/vite-plugin-svelte/src/plugins/setup-optimizer.js
+++ b/packages/vite-plugin-svelte/src/plugins/setup-optimizer.js
@@ -8,6 +8,7 @@
 import fs from 'node:fs/promises';
 import path from 'node:path';
 import * as svelte from 'svelte/compiler';
+import { parseSync } from 'vite';
 import { log } from '../utils/log.js';
 import { toRollupError } from '../utils/error.js';
 import { SVELTE_IMPORTS } from '../utils/constants.js';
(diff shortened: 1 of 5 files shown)


packages/vite-plugin-svelte/src/plugins/setup-optimizer.js
```diff
@@ -98,7 +98,21 @@ function rolldownOptimizerPlugin(api, consumer, components) {
 			delete plugin.buildStart;
 			delete plugin.transform;
 			delete plugin.buildEnd;
+			if (components) {
+				plugin.load = {
+					filter: { id: includeRe },
+					async handler(filename) {
+						try {
+							const code = await fs.readFile(filename.replace(/[?#].*$/s, ''), 'utf8');
+							return await compileFn(api.options, { filename, code }, generate, statsCollection);
```
teemingc: This is kind of expensive to do without even checking if there are exported snippets. I don't think we have to do this since it's not done for `lang="ts"` code blocks currently. A better solution might be to avoid the error during the scan by some other means.

teemingc: I wonder if treating the script as typescript during scanning so that we get the same behaviour would actually solve this or if we're just skirting around the problem
```js
      plugin.transform = {
        filter: { id: includeRe },
        handler(code) {
          return { code, moduleType: "ts" };
        },
      };
```

jkbz64: Would transforming w/ `filter: { id: includeRe, moduleType: ['js'], code: /\bexport\b/ },` and stubbing out snippet names be acceptable? e.g vite's copy ending up with
```js
export { snippet };
const snippet = null;
```

Treating as `ts` feels weird too... Honestly I don't like any of the solutions, feel free to supersede with the `ts` approach or any other idea if this is still too expensive.

teemingc: Yeah, the `ts` solution doesn't work. We could try stubbing them and if that doesn't work too, maybe compiling isn't so bad? I don't know if the next transform will be cached the transform hook would've run for that module ID already.

PR #1385, opened 2026-08-05
fix: configure api when plugin is used replaced with `applyToEnvironment`

paoloricciuti:
This PR fixes a problem when using the svelte plugin with `applyToEnvironment` or `perEnvironmentPlugin` (which is just a thin wrapper around `applyToEnvironment`).

The issue is that when there's multiple environment, and you use `applyToEnvironment` the `configResolved` hook is invoked before `applyToEnvironment` and since we assign to the `api` object only in `configResolved` the other plugins that try to use `api` fails.

The solution is to retry to load the config inside `buildStart` (which despite the name is also invoked in dev) unless the api has already been configured. There's a problem, tho: other plugins also relied on `configResolved`. I tried different solutions like always using `buildStart` instead of `configResolved` (this failed when not using `applyToEnvironment`), using a getter/setter for `options` but none felt better than what I have here which is adding a callback registry to the `api` object so that plugins can register there instead of using `configResolved`.

I tried and it seems to work well but please let me know if you have better ideas.

### DIFF
--- a/.changeset/upset-melons-fix.md
+++ b/.changeset/upset-melons-fix.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/vite-plugin-svelte': patch
+---
+
+fix: configure api when plugin is used replaced with `applyToEnvironment`
--- a/packages/vite-plugin-svelte/__tests__/environment-api.spec.js
+++ b/packages/vite-plugin-svelte/__tests__/environment-api.spec.js
@@ -1,0 +1,36 @@
+import fs from 'node:fs/promises';
+import path from 'node:path';
+import { fileURLToPath } from 'node:url';
+import { afterEach, describe, expect, it } from 'vitest';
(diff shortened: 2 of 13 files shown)


teemingc:
After looking deep into this, I'm convinced we're better off lobbying for "Environment plugins resolution may be moved before configResolved in the future." to come quickly (mentioned in https://vite.dev/guide/api-environment-plugins#per-environment-plugins-using-the-applytoenvironment-hook).

Right now, _all_ global hooks (config, configResolved, etc.) are being skipped when using the `perEnvironmentPlugin` helper. So, this wouldn't work even if you refactored all the other hook logic that's being skipped because we need to modify the applied Vite config in the `configEnvironment` hook (which is also being skipped because it runs before `configResolved`).

paoloricciuti:
Mmm maybe you are right...it's a bit of a shame that users can't actually use this but let's open an issue on vite and let's see what they say

PR #1386, opened 2026-08-06
feat: pass environment to `dynamicCompileOptions`

paoloricciuti:
We closed #1318 but the suggestion we gave to use `perEnvironmentPlugin` didn't work because of the current vite setup that doesn't invoke (or invoke too late) `resolveConfig`.

I've [opened a bug in vite](https://github.com/vitejs/vite/issues/23176) to see if it's solvable, but in the meantime we can pass the environment to dynamic compile option to solve it.

### DIFF
--- a/packages/vite-plugin-svelte/src/plugins/setup-optimizer.js
+++ b/packages/vite-plugin-svelte/src/plugins/setup-optimizer.js
@@ -43,8 +43,8 @@ export function setupOptimizer(api) {
 			// Add optimizer plugins to prebundle Svelte files.
 			optimizeDeps.rolldownOptions = {
 				plugins: [
-					rolldownOptimizerPlugin(api, consumer, true),
-					rolldownOptimizerPlugin(api, consumer, false)
+					rolldownOptimizerPlugin(api, name, consumer, true),
+					rolldownOptimizerPlugin(api, name, consumer, false)
 				]
 			};
 
@@ -71,13 +71,13 @@ export function setupOptimizer(api) {
 
 /**
  * @param {import('../types/plugin-api.d.ts').PluginAPI} api
+ * @param {string} environmentName
  * @param {'server'|'client'} consumer
  * @param {boolean} components
  * @return {Rolldown.Plugin}
  */
-function rolldownOptimizerPlugin(api, consumer, components) {
+function rolldownOptimizerPlugin(api, environmentName, consumer, components) {
 	const name = components ? optimizeSveltePluginName : optimizeSvelteModulePluginName;
-	const compileFn = components ? compileSvelte : compileSvelteModule;
 	const statsName = components ? 'prebundle library components' : 'prebundle library modules';
 	const includeRe = components ? /^[^?#]+\.svelte(?:[?#]|$)/ : /^[^?#]+\.svelte\.[jt]s(?:[?#]|$)/;
 	const generate = consumer === 'server' ? 'server' : 'client';
@@ -107,7 +107,22 @@ function rolldownOptimizerPlugin(api, consumer, components) {
 				 */
 				async handler(code, filename) {
 					try {
-						return await compileFn(api.options, { filename, code }, generate, statsCollection);
+						if (components) {
+							const environment = api.options.server?.environments[environmentName];
+							return await compileSvelte(
+								api.options,
+								{ filename, code },
+								generate,
+								environment,
+								statsCollection
+							);
+						}
+						return await compileSvelteModule(
+							api.options,
+							{ filename, code },
+							generate,
+							statsCollection
+						);
 					} catch (e) {
 						throw toRollupError(e, api.options);
 					}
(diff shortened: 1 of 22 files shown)


packages/vite-plugin-svelte/src/plugins/setup-optimizer.js
```diff
@@ -107,7 +107,22 @@ function rolldownOptimizerPlugin(api, consumer, components) {
...
 				async handler(code, filename) {
 					try {
-						return await compileFn(api.options, { filename, code }, generate, statsCollection);
+						if (components) {
+							const environment = api.options.server?.environments[environmentName];
+							return await compileSvelte(
+								api.options,
+								{ filename, code },
+								generate,
+								environment,
```
teemingc:
```suggestion
							return await compileSvelte(
								api.options,
								{ filename, code },
								generate,
								this.environment,
```
This way the environment is always available; it uses the environment the hook is currently running in. In contrast, `server` is only available during development.

packages/vite-plugin-svelte/src/public.d.ts
```diff
@@ -86,6 +87,7 @@ export interface PluginOptions {
 		filename: string;
 		code: string;
 		compileOptions: Partial<CompileOptions>;
+		environment?: Environment;
```
teemingc:
```suggestion
		environment: Environment;
```

This should more or less always be available.

packages/vite-plugin-svelte/src/public.d.ts
```diff
@@ -69,6 +69,7 @@ export interface PluginOptions {
 	 * `data.filename` - The file to be compiled
 	 * `data.code` - The preprocessed Svelte code
 	 * `data.compileOptions` - The current compiler options
+	 * `data.environment` - The current Vite environment, if available
```
teemingc:
```suggestion
	 * `data.environment` - The current Vite environment
```

packages/vite-plugin-svelte/src/plugins/setup-optimizer.js
```diff
@@ -131,10 +146,11 @@ function rolldownOptimizerPlugin(api, consumer, components) {
  * @param {ResolvedOptions} options
  * @param {{ filename: string, code: string }} input
  * @param {'client'|'server'} generate
+ * @param {import('vite').Environment | undefined} environment
```
teemingc:
```suggestion
 * @param {import('vite').Environment} environment
```

teemingc (changes requested):

packages/vite-plugin-svelte/src/plugins/setup-optimizer.js
```diff
@@ -71,13 +71,13 @@ export function setupOptimizer(api) {
 
 /**
  * @param {import('../types/plugin-api.d.ts').PluginAPI} api
+ * @param {string} environmentName
```
teemingc, line 74:
```suggestion
```

Probably don't need this now since we can access the environment from the plugin hook

packages/vite-plugin-svelte/src/plugins/setup-optimizer.js
paoloricciuti: Mmm I did this initially but I think the environment there is a different shape from the one you usually get..let me try

teemingc: Yeah, it's a shame Vite information isn't available when the plugin runs as part of `optimizeDeps`. It's probably better that we get rid of the `undefined` part of the types since that comes from `ResolvedConfig` and isn't accurate. The Vite dev server would have already been available by the time the `transform` hook hits so we would definitely have the environment instance.

paoloricciuti: If you are sure we can always provide that and it will always be defined than yeah I'll remove the undefined

teemingc: Yeah, the order of hooks are:
1. `config`
2. `configEnvironment`
3. `configResolved`
4. `configureServer` <-- development server is created here
5. Rolldown stuff

paoloricciuti: Allright then removing it!

teemingc (approved):

packages/vite-plugin-svelte/src/plugins/setup-optimizer.js
teemingc, line 231: A little off-topic but it's a bit peculiar that the dynamicCompileOptions don't also apply here. Should they?

paoloricciuti: Yeah I was puzzled as well but I'd say we can move this to a different PR?

teemingc: Yeah definitely a different one

